PR #1, opened 2015-11-09
Testing how Pull Requests Work

SgtBears:
I have added Rainforest to all the Jungle Events and have also standardized all events to obsolete at the exact start of an era rathar than a few technologies ahead or behind.


dguenms:
It worked! Please request the merge for develop though, I will not merge into master and use that branch only for releases (and maybe larger features within a major version). I can then accept it right after 1.13 is released.

PR #2, opened 2015-11-09
Event Trigger Fixes

SgtBears:
Like This?


dguenms:
Yes, thanks. Will merge right after release.

PR #6, opened 2015-11-22
Senegal and Mozambique

epgquota:
In 1700 AD scenario, added Saint-Louis du Sénégal for France and renamed Laurenço Marques to Moçambique for Portugal.


dguenms:
From my research, Mozambique is only the name of the island and not the city on it. Maybe Sao Sebastiao is better? It would also be nice to have an industrial era transition to Laurenco Marques so Maputo is a possibility, you can mirror Istanbul for this.

epgquota:
Evidence I have seen suggests that, like Venice or Gorée, the island and the settlement aren't distinguished by name. I used sources like the 1911 Encyclopedia Britannica (http://www.theodora.com/encyclopedia/p2/portuguese_east_africa.html : "On one of these islands is Mozambique, and immediately north of that port is Conducia Bay") and the Catholic Encyclopedia (http://www.newadvent.org/cathen/12311a.htm : "For purposes of justice the province is divided into seven comarcas and the town of Mozambique has a Tribunal of Second Instance composed of three judges"), and although these are relatively modern and English-language sources, I haven't seen anything to countermand them. My only small concern about Lourenco Marques/Maputo is that the city is actually located south of the Limpopo river (i.e. right beside South Africa), while the current city site is east of the Zambezi.

dguenms:
Okay, looks like your research has been more thorough :)

I think the geographic inaccuracy is preferable to the long term anachronism of the name, though.

epgquota:
OK I will go ahead with Istanbul-type change.

dguenms:
:+1:

PR #8, opened 2015-11-23
Corrected Misspellings

SgtBears:
Corrected some more misspellings in the new corporation advisor. There is no need for the other languages to be there when they are untranslated.


dguenms:
I don't see any misspellings fixed in the diff?

Having the XML tags for other languages already there is generally preferable, it makes translating them easier later on.

SgtBears:
Caribbean, Sub-Saharan, Customs Houses and Competition were misspelled in all of them for starters. (ignore the non corporation text keys I didn't change any of them so I don't know why they are showing up as changed)

Way back when I cleaned the files for translating I planned to use transifix which requires removing all incorrect and placeholder translator tags, so I did a dif with the original files and then the RFC files to see what tags had been changed and therefore what translations had become outdated and then deleted them. (I am so sick of looking at text key diffs)

Under your system translators have to check every language tag cause some of them are outdated or placeholders. Under mine translators only have to check text keys that don't have the language tag because I have removed all incorrect or placeholder language tags. My system does have the disadvantage that I have deleted some texts that only needed a little bit of tweaking to update them but there aren't many of them.

dguenms:
Oh, so that is how Transifex handles things (are you the owner of the existing project there by the way). Makes sense to delete the tags then.

And now I see what changes you made, somehow I expected github to highlight changes between lines like WinMerge does. Looks good.

SgtBears:
Yeah I created that transifex project when I created the translation thread ages ago on Civfanatics.

PR #9, opened 2015-11-23
Update CIV4EventTriggerInfos.xml

SgtBears:
Updated the event triggers again.


dguenms:
This is because Democracy has moved to the Industrial era, correct? I am thinking about reverting that again, so it might not be necessary anymore ...

SgtBears:
Yes I only updated because Democracy became part of the Industrial Era.

PR #10, opened 2015-11-23
Building Ordering

SgtBears:
Has no changes except ordering the buildings. All unique buildings are under the base buildings, the wonders are sorted by when they appear in the tech tree.


CvGameCoreDLL/CvRhyes.h
```diff
@@ -11,10 +11,10 @@ typedef list<char*> LISTCHAR;
 
 #define MAX_COM_SHRINE			(20)
 
-#define GREATPALACE				(1)
-#define SUMMERPALACE			(1)
-//#define VERSAILLES				(2)
-#define FORBIDDENPALACE			(2)
+#define GREATPALACE				(GC.getInfoTypeForString("BUILDING_GREAT_PALACE"))
```
dguenms, line 14: We should not use getInfoTypeForString here ... defines do not store constants, they only do code replacement. In other words, the expensive string operation is called whenever the constant is queried which is not what we want.

dguenms:
Besides that code part, everything looks good. How much in game checking did you do to see if all constants match up?

PR #12, opened 2015-11-24
Minor Python fixes + some city additions

temujin2k:
Hello!

I've made some minor changes to the Python code regarding the behavior of some things. These may be found in [this commit in my fork](https://github.com/bnxkpc/Dawn-of-Civilization/commit/0aeb7990f553da5c6dc2b919751f5716f444bd7b). I'm not totally sure if the original behavior was intended and I wanted to see if the changes I made are better.

Also, I've added a few cities, three spawning late and three preplaced in the 600 AD scenario. I was aiming to populate certain areas a bit more while aiming for some historicity too.

Any suggestions are welcome, although I might be able to address them by at least tomorrow night (UTC +08:00). Thank you! :smile:


dguenms:
Sorry for the late reply.

I don't really want to have more initial/spawning cities right now. As far as the birth in capital mechanism is concerned, this is intentional (civs that are not labeled as conditional are supposed to erase cities previously present on the tile). This is important especially for one tile away cities - for some civs it just doesn't make sense to flip those.

Haven't looked into the Tokyo capital move fix yet, will follow up on that later.

By the way, you can always submit several pull requests for distinct features, that makes discussing and integrating them easier and less confusing.

dguenms:
Incorporated the Japanese capital transition fix in a separate commit.

temujin2k:
Hi!

Sorry I wasn't able to tend to this right away; I was busy preparing for my final exams this week. Thank you, by the way for incorporating the Japanese capital transition fix! It's an honor :)

Anyways, about the Bogota hill tile, I've read way back in the past that the Andes mountains should open up with the Inca conqueror event. I looked into the code in Resources.py in the past, and what was placed there was that it opens up in 1700 AD. Is it really intended that it open up in 1700 AD or should it be during the conqueror event?

That is my only concern for now. I'll just finish my last final exam and maybe a coding requirement for one of my classes and then I'll look for more changes I can think of. Thank you! :)
- @h0spitall3rz in CFC

dguenms:
The pass that opens up with the conquerors is in Chile / southern Andes. Maybe the conqueror event is indeed better though, that could allow for an earlier Bogota. It seems I didn't read your changes in enough detail; it looked like you were only removing the peak to hill code without replacement.

Good luck with your exams in the meantime!

temujin2k:
Ohhh. I never knew that one was referring to the southern Andes. Well, opening up Bogota gives everyone an alternate route into the Andes. And considering that Bogota was founded a bit early (based from what I read), I think the Spanish should be able to settle there earlier. I moved the code from Resources.py to RiseAndFall.py since the conqueror event code is also there.

Thank you! The last exam should be done by noon today, so I can start giving some more time to DoC. :+1:

PR #21, opened 2015-12-08
Platybuilder

MeuhV1:
Fixed a small bug in the PlayerData screen
Added a feature to enable/disable secondary civs via the WB (PlayerData screen)

Sorry about the messy changelog, I just can't make it clean. I just don't understand Git well enough yet.


dguenms:
Can you make the commit history a question in either the Git or modmodding thread? That is, what exactly did you do? It makes it hard to manually squash the history every time, and it should only be a rebase for you.

MeuhV1:
I thought I rebased it, but it seems something went wrong there.

What do you exaclty mean with commit history?

dguenms:
Oh, I just meant the general topic. If the rebase went through, maybe you need to revert your commits first. I think it's easier to discuss that on the forum.

dguenms:
Okay, this apparently easier than I thought with TortoiseGit. Look at TortoiseGit -> Show log. The list there should show all commits including the repeating/test/fix commits we want to get rid of. You can select multiple commits with the Ctrl key. Afterwards right clicking should give you the option to combine the selected commits.

Clean up your develop branch this way. Not sure if rebasing from develop will fix this branch then, otherwise you have to do the same again here.

MeuhV1:
I don't see an option combine commits or something similar when I right click after selecting multiple commits.

dguenms:
Not even when selecting them pairwise?

dguenms:
Squashed commits and merged.

PR #34, opened 2015-12-16
Text Changes

SgtBears:
These are my text changes, I have improved (hopefully) many text keys, corrected all pedia and strategy texts and removed all incorrect or outdated translations

I have included all of the latest text key updates, however the organised text files I gave you earlier will be outdated.

I also included the program I used to sort the text keys.


SgtBears:
These Texts are outdated, so don't merge them until I update them.

dguenms:
Please wait until I can comment before you invest further work into this.

dguenms:
Okay, it's completely impossible to review changes if they come in such large increments. That's 20000 lines! I know most of that comes from reordering between files, but there is no way for me to find the actual changes between the reordering.

Second, aggregating too many text keys into only a few files does more harm than good. It only becomes harder to find the file they are actually in when trying to modify them. For example, the files from external mods and dynamic names per civ should definitely stay as they are.

I agree that currently there are too many text files in that folder, and subdirectories are unfortunately not an option. However, I can move the external mod text and dynamic name texts into modules instead.

So I suggest the following:
1. You wait until I move the external/dynamic name files elsewhere (you can continue working on the rest in the meantime of course)
2. pull request for the changed text strings
3. pull request for the changed text keys
4. pull request for the removed missing translations
5. only one last pull request that does nothing but reordering and rearranging between files

MeuhV1:
Are subdirectories not an option because it's technically not possible or you just don't want them? I'm pretty sure I was able to use subdirectories for text files in another mod.

dguenms:
I just assumed that it would not work, let me try that first. It's definitely the preferable solution.

dguenms:
It does work, I have moved external mod files and dynamic names into subdirectories with the latest commit.

PR #42, opened 2015-12-16
Civilization Xml Cleanup

SgtBears:
CIV4CivilizationInfos.xml:  Removed Rhye Notes
                Moves all City names to TEXT_KEY (I will probably change this later to moves all city names out of TEXT_KEY(
                Removes Leaders that never appear from their civilization
                Removes Commented Out Babylonian Garden
                Removes Starting Techs
                Removed some pointless spaces, tabs and lines
                Makes Celts, Barbarians, Seljuks, Independents and Natives unable to build national wonders (except palace for seljuks), National Securities (so that they can't pick up spies easier which you don't want from independents), Scouts and Explorers (You don't want their recon units getting in the way), Missionaries and Persecutors (You don't want them playing with religions), Slaves (They get Native Slaves that cannot do anything)
                Removes Unique Units and Buildings from Native Americans (Civilopedia reasons), Zulus (Civilopedia reasons), Sumerians (Civilopedia reasons) and Natives (So that natives in the americas don't build Impis) But add Native Slaves to all of them.

CIV4LeaderHeadInfos.xml:    Removed Rhye Notes
                Removed some pointless Notes
                Removed some pointless spaces, tabs and lines
                Corrected some Text Keys
                Corrected some Notes


MeuhV1:
Do the civs that start in 3000 BC (Egypt, Babylon, China, Harappa) still have some techs when you start a game? IIRC, they get their techs because they are defined here. So they shouldn't be removed.

SgtBears:
Maybe, but their starting technologies are also defined in the 3000BC map. Which takes precedence over civilization xml file.

MeuhV1:
I forgot about that. Then it's fine.

dguenms:
Is this dependent on the text key changes?

Did you change anything about LHs? Reordered them? The diff is once again too large to view in the browser UI so it is hard to tell, but it looks like you changed some AI things like improvement or unit AI preferences and memory decay.

Also in some cases you seem to have changed comments from "predefined personality" to a specific AI personality, why?

As for the city names, I don't think it makes sense to move them to text keys. Unlike BtS, in DoC city names do not depend on game language, but on the controlling civilization, so there is not reason to use keys. And doing so would make the civ defines less readable and bloat the text files.

SgtBears:
Yes it is dependent of text key changes.

I changed some spaces in front of tags into tabs to match the rest of the file. This creates an effect which looks like a whole host of changes but I haven't changed any Leader effects.

The changes from predefined personality to specific ai personality. I saw no reason for the predefined leader personality to go unnamed. So I searched through the files for the nearest match among all the original civ leaders.

I reversed the text key city name changes and changed some city text keys to pure text. (I will do the rest some other time)

I forgot to mention I have added some canadian unqiue city names (originally created by I think it was DC123456789) and have added all turkish city names to the Seljuks (in case they ever gain the ability to found cities) and added all harappan citynamemanager names to the harappan city list.

SgtBears:
Changed all city text keys to pure text. Removed Harappa, Gaul, Polynesia and Phoenicia from the barbarian and independent city lists since all of these are civilizations in-game. I know I don't need to but just in case these city names are ever used.

dguenms:
I think it is rather confusing, the purpose of the comments in the LH section is to signify that an exact copy is being used. If it is not an exact copy (every time a leader from the database came with their own personality), the comment should reflect that.

SgtBears:
Okay, I have reversed the Leaderhead comments.

SgtBears:
Updated.

dguenms:
Merged.

PR #49, opened 2015-12-28
Small fixes (various)

MeuhV1:
Fixed missing TXT key for Moors DCN
Fixed TXT key for Polynesian DCN
Fixed error in RFCUtils at makeUnit
Fixed bug when renaming city when being Harappa
Fixed Colombian leader
Fixed Civic screen resolution for lower screen resolutions (It's slighty smaller as well on higher resolutions, but it's still very well readable)
Pillaging with Corsair counts towards UHV


Assets/Python/CvRFCEventHandler.py
```diff
@@ -453,8 +453,9 @@ def onCityGrowth(self, argsList):
 	def onUnitPillage(self, argsList):
 		unit, iImprovement, iRoute, iPlayer, iGold = argsList
 		
-		if iPlayer == iVikings and iGold > 0 and iImprovement != -1 and iGold < 1000:
-			vic.onUnitPillage(iPlayer, iGold)
+		iUnit = unit.getUnitType()
+		if (iPlayer == iVikings or iPlayer == iMoors) and iGold > 0 and iImprovement != -1 and iGold < 1000:
```
dguenms, line 446: The Moors do not have a pillage related goal, what is the idea behind this?

Assets/Python/DynamicCivs.py
```diff
@@ -1784,6 +1784,10 @@ def leader(iPlayer):
 	elif iPlayer == iRome:
 		if not bEmpire and not bCityStates: return iAugustus
 		
+	elif iPlayer == iMaya:
```
dguenms, line 1788: Good catch.

Assets/Python/RFCUtils.py
```diff
@@ -282,7 +282,7 @@ def makeUnit(self, iUnit, iPlayer, tCoords, iNum, sAdj="", iExp = 0): #by LOQ
 			if iExp > 0:
 				unit.changeExperience(iExp, 100, False, False, False)
 				
-		return unit
+		#return unit
```
dguenms, line 288: Yes, it's probably better to remove that.

Assets/Python/Victory.py
```diff
@@ -1848,12 +1848,16 @@ def onGreatPersonBorn(iPlayer, unit):
...
 					
-def onUnitPillage(iPlayer, iGold):
+def onUnitPillage(iPlayer, iGold, iUnit):
 
 	# third Viking goal: acquire 3000 gold by pillaging, conquering cities and sinking ships by 1500 AD
 	if iPlayer == iVikings:
 		if isPossible(iVikings, 2):
 			sd.changeVikingGold(iGold)
+			
+	elif iPlayer == iMoors:
```
dguenms, line 1859: Oh, I see. The Moorish goal is intended to only count sunk ships, is the goal phrased differently?

Assets/XML/Text/DynamicNames/DynamicNames_Moors.xml
```diff
@@ -40,4 +40,8 @@
 		<Tag>TXT_KEY_CIV_ARABIAN_NAME_ARABIA</Tag>
 		<English>Hejaz</English>
 	</TEXT>
+	<TEXT>
+		<Tag>TXT_KEY_CIV_MOORS_EMIRATE_OF</Tag>
```
dguenms: Should be Emirate of %s1.

Assets/XML/Units/CIV4PromotionInfos.xml
```diff
@@ -4184,7 +4184,7 @@
 			<bShiftDown>0</bShiftDown>
 			<bCtrlDown>0</bCtrlDown>
 			<iHotKeyPriority>0</iHotKeyPriority>
-			<Button>,Art/Interface/Buttons/Promotions/Logistics1.dds</Button>
```
dguenms, line 4187: Are you sure this is necessary? The buttons do work in the game don't they? I think I know of crashes/loading problems without the leading comma in some systems. (Same for below.)

dguenms:
The branch is out of synch (which is probably my fault for delaying the review for so long), can you do a rebase after addressing the other comments so I can merge this?

MeuhV1:
I rebased the branch.

The TXT key for the Moors were fixed by sgt bears in the meantime, so I removed it.

IIRC, the comma in fromt on the buttons cause a CTD if that button appears in the citybar. As you can't build promotions, this will not appear. So it isn't really necessary. But this avoids any risk. (Maybe the comma can cause a CTD for the same reason in another situation we didn't encounter yet)

I just think that pillaging  sea improvements is also a part of pirating. But if you don't think it is suitable, feel free to ignore it. (It was more of a suggestion than a real fix)

dguenms:
Okay, I looked into the comma stuff again and you're right. This is the safer way of doing it.

I'll think about the Moorish UHV goal. In the very least, the description should probably be more specific.

PR #50, opened 2015-12-29
Platybuilder update New fork

MeuhV1:
NOTE:
As of typing this note, I have already rebased this branch. But I still have to update some code because of the changes. So it's not ready to be merged yet.

Old changes:
- Fixed typo (fortifying was written without a t)
- Spawn locations have the color of the civ instead of magenta or black
- Stability overlay like SoI
- Stability editor tools and export *

Changes done between:
- Updated stability overlay so it has a tooltip (requires recompiling DLL) **
- Replaced the regenerate map button (wasn't used anyway) with a button that does convertsave.py -(so people don't need to use the python console)
- Fixed bug when stabmap "paint" tool is used while a nonplayable civ is selected
- A mode that shows the CNM when hovering over a tile
- I tried to keep the functional part in one, new file. (Mapdrawer.py (I wasn't very creative with the name)) The changes in the other platybuilderfiles are graphical changes. (To make buttons appears for example)
  With the recent constant to python changes, I may have to redo some parts. (like tTL and tBR)

*\* edead created a new controltype for the overlay, which requires recompiling the DLL. I thought I could get around this by triggering the toggle in python. It works, but it doesn't create a tooltip when hoovering over the button. I added the 3 GameCore files, but left the recompiling over to you. You also have to uncomment line 5592 in CvMainInterface.py to make it work, and comment out/remove line 5593. (I disabled it because I got a pythex because the new controltype wasn't in my DLL yet)


dguenms:
Not sure if you've noticed, this branch is also currently in conflict.

MeuhV1:
Yeah, I noticed. I will fix all the conflicts in the progress of updating.

PR #52, opened 2016-01-01
Other Unit Xml Cleanup

SgtBears:
Dguenms commented:
I don't really understand what you are talking about with regards to unique unit promotions. They work fine as they are.

In BTS all unique units with promotions that they normally cannot access are granted access to all promotions in that promotion chain. i.e. BTS Samurai starts with Drill I which melee units cannot normally access and the samurai can also access Drill II, Drill III and Drill IV.
What I have done is extend this principle to all unique units in DOC and removed it from the Drill line since the samurai no longer gets Drill I.


dguenms:
I don't think the promotion change is necessary. But it is apparently not part of this pull request anyway?

SgtBears:
It was included but I have removed it. If you don't want the units to gain access to a promotion line that the unit doesn't normally access because of the free promotions (a la BTS) then the bamboo bowman and the colombian rifleman should gain direct bonuses rather than the unique promotions that they get for free.

If you agree I will upload the direct bonuses to 2nd unit xml changes.

dguenms:
I think it's fine as it is.

dguenms:
Merged.

PR #54, opened 2016-01-01
Second Unit Xml Cleanup

SgtBears:
All the other changes I mentioned in the old merge.


dguenms:
Which pull request do you mean? (Link is sufficient)

The branch is also out of date as it is.

PR #56, opened 2016-01-02
New cnm

hugh-mann:
Modification to CityNameManager.py that moves the CityFoundingMap data into their own folder and builds dFoundMaps with CityFoundingMapsWrapper.py. Individual founding maps are found in CityFoundingMaps/m<Language>.py as their own modules and are eventually folded into CityNameManager.py. This change isolates the data into their own files, helping separate code from data.
- Makes modding easier by separating data and code and makes code easier to maintain
- Better generalizes the process of adding a city founding map for integrating new civilizations
- Makes adding additional founding map languages easier.


dguenms:
Shouldn't the rename dicts and the identifiers dict be equally refactored out, to complete the separation of data and code?

hugh-mann:
I can do both; this was initially just tweaking the language name dictionaries, but I can refactor out both both of those too. Gimme a week or two to do it; this time of year is busy for me.

dguenms:
Don't worry, there is no hurry at all with this.

dguenms:
I'm closing the request for now, can reopen if desired.

PR #60, opened 2016-01-15
Text Fixes

SgtBears:
This only includes Actions, Buildings and Projects. I will be adding to this alot but merge whenever you want. I will just create a new merge.

This only changes tag types, removes old or faulty translations, adds new translations and fixes strategy and pedia texts. I will include other text changes in a later merge.


dguenms:
Merged.

PR #61, opened 2016-01-16
Platybuilder

MeuhV1:
Due to merging issues, I created a new branch
All previous updates updated to new Consts to python

Fixed typo (fortifying was written without a t)
Spawn locations have the color of the civ instead of magenta or black
Stability overlay like SoI (requires recompiling DLL) **
Stability editor tools (core + settlervalue)
Export tool (only required for permanent changes)
Replaced the regenerate map button (wasn't used anyway) with a button that does convertsave.py (so people don't need to use the python console)
A mode that shows the CNM when hovering over a tile
Civ specific modifiers editor in PlayerData screen

Flipzone editing is temporarily disabled

*\* edead created a new controltype for the overlay, which requires recompiling the DLL. I thought I could get around this by triggering the toggle in python. It works, but it doesn't create a tooltip when hoovering over the button. I added the 3 GameCore files, but left the recompiling over to you. You also have to uncomment line 5592 in CvMainInterface.py to make it work, and comment out/remove line 5593. (I disabled it because I got a pythex because the new controltype wasn't in my DLL yet)


dguenms:
Did you reopen this recently?

Although it's possible to merge, doing so would revert a couple of changes from my most recent commits. So another rebase is required first.

MeuhV1:
I noticed it. I reverted it back to the most recent commits. (And included 2 files I forgot to commit)

dguenms:
Did you make those changes by hand again? That will just leave us with a messy history.

Simply merge in my own commits, or rebase to develop.

MeuhV1:
Redid the changes. Now it's only 1 commit.

I finally learned how to get back to an old version and delete the remote commits. So most of my merging issues can be solved this way.

dguenms:
Looks good!

Unfortunately not everything seems to work even with the DLL compiled. I'm getting Python exceptions when trying to edit settler values for example.

If you want to continue working with a compiled DLL, I have committed it to the MeuhV1-Playbuilder branch.

MeuhV1:
I don't see the new GameCore. (But that is only required for the SoI stability overlay)
EDIT: I see where it is.

In what screen did you get the python exceptions?
EDIT: Was the python exception the stability overlay? (I told you do do that ;)) Are there any other python exceptions?

dguenms:
Sorry for not being specific, I really should know better :) I'm going to make a forum thread for this, it sounds like there is more discussion to be had here. Not that I can't merge this in the meantime (as soon as the exceptions are fixed), we can iterate on it later.

MeuhV1:
I got the new DLL and suddenly, I got the error too. However, when I look at the changed files, the code is correct. Somehow, some brackets were removed.

dguenms:
I did remove the brackets to conform to the method API (DLL methods do not accept tuples). That was to fix a different error.

dguenms:
Merged for the moment. You can continue committing to this branch and just open another pull request for the same branch if you think another feature is complete.

PR #63, opened 2016-01-17
Unit Xml Fixes

SgtBears:
Removed all ProductionTraits mentions since they weren't used anyway.
Removed Woodsman 1 from Brazilian worker since they cannot defend they don't use its abilities.
Added GraphicalOnly to Indian Fast Worker, French Musketeer, Cidainh, Camel Gunner,
Increases Mongol Knight CollateralDamageMaxUnits from 5 - 6 to match all other collateral damage units.
Changed Seljuk Ghulam Knight tech requirements from Horseback riding and Archery to Feudalism and Horseback riding to match the knight. The Seljuks start with all mentioned technologies so there is no practical difference.
Reduces Cannon and French Cannon AdvancedStartCost from 120 to 100 to match BTS and all other units.
Adds ignorebuildingdefense to the spanish conquistador to match all other gunpowder units.
Makes Stealth Destroyers able to see Stealth Destroyers. Before no unit could see stealth destroyers and therefore couldn't attack them.
Makes all Great People Nuke Immune.
Makes all Great Prophets able to build Orthodox Shrine. The orthodox shrine never has to be built because it is created automatically but having them able to build the shrine means that if a bug occurs players can still create it.


dguenms:
Merged.

PR #65, opened 2016-01-18
Removing All Unused Tags

SgtBears:
This makes all tags in the Units.xml and Buildings.xml optional and then removes all empty or default tags from their files. I would like to extend this to a few other xml files like promotions.xml but I just want to see what you think of this.


dguenms:
No, this is great. I think another file where this is useful are Civics.

SgtBears:
Fixed Civics and Promotions.

dguenms:
Also in.

PR #66, opened 2016-01-19
Building and Unit Fixes

SgtBears:
Adds some relevant unneeded tags to the Building Unit Xml.
Changes Native_America_totem obsoletetech from TECH_CALENDAR to TECH_DIVINE_RIGHT
Removed ReligionTypeRELIGION_ISLAM/ReligionType from the Madrassa
Adds UnitClassUpgradeTypeUNITCLASS_CUIRASSIER/UnitClassUpgradeType to the Keshik and the Chang Suek. Since they both upgrade to cavalry the player can still build them after cuirassiers become available but can upgrade them if the player would prefer.
Reduces the keshik collateraldamagemaxunits from 6 - 5. It was a mistake I made.
Adds FlankingStrikeUnitClassUNITCLASS_BOMBARD/FlankingStrikeUnitClass to the Argentine Cavalry to match the cavalry unit.

### DIFF
--- a/Assets/XML/Buildings/CIV4BuildingInfos.xml
+++ b/Assets/XML/Buildings/CIV4BuildingInfos.xml
@@ -230,6 +230,7 @@
 				</SpecialistCount>
 			</SpecialistCounts>
 			<ConstructSound>AS2D_BUILD_CASTLE</ConstructSound>
+			<BonusProductionModifiers/>
 			<UnitCombatFreeExperiences>
 				<UnitCombatFreeExperience>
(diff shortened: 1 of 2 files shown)


dguenms:
Can you restore the religion type for Madrassas? It enables them to benefit from wonders like the Spiral Minaret, which is desirable imo.

Keshiks and Chang Sueks deliberately do not upgrade to Cuirassiers. For the former, it is to force Mongolia into an era of technological inferiority as technology progresses. For the latter, it is to keep them relevant for Thailand a bit longer.

Anything else is ready to merge.

SgtBears:
Fixed

dguenms:
Merged.

PR #68, opened 2016-01-20
Platybuilder update

MeuhV1:
NOTE:
The pull request contains 1 DLL file which has to be recompiled.
(Required to change the color of Core tiles in the plotinfo-tooltip to match the new cyan color)

General:
getForeignCores function ignores peaks and water tiles
Fixed double wondermovie bug (not platybuilder specific, but anyway ;))
Added colors to stability overlay tooltip
Fixed PlayerData religion bugs
Removed convertsave button
Fixed 2 missing tiles in the Chinese warmap

Infoscreen:
Removed mapeditor functions from the infoscreen. It's now only a viewer
Adjust plottable so the bottom matches the bottom of the infotable in infoscreen (aesthetics)
The map in the is always visible in all modes in the Infoscreen (even if no items are present)

PlotInfo:
Added option in Plotinfo to have keyboard input to change settlervalue

MapEditor:
Flipzone mode is viewonly
Changed color for Core to cyan (more distinction with historical area)
Stability overlay has colored hatched lines like settlers
CNM isn't a seperate mode anymore. The cityname will always show on top of the screen with the settlervalue (only in all mapeditor modes)
Combined core and settlermap overlay
Updated Revert and Export panel, it will revert/export only 1 map type for only 1 civ (selected civ)
Added switchreborn button in revert/export panel (only for the civs that can switch)
Editing water and peak tiles will be ignored (if boolean is on. On by default)

Warmap editor:
Works similar to flipzone editor
Warvalue shown instead of settlervalue for "hoovering" tile (all other map modes show the settlervalue)

Export:
New location (in the mod folder itself. Autofinds the mod folder)
Updated output so it can be copy-pasted directly in the designated files (see the Egypt examples. Don't use them though, they contain random map changes)
New boolean that automatically sets the settlervalue and warvalue of core tiles to respectively 500 and 8 at least if turned on when exporting maps (on by default)


dguenms:
I'll give it a look later today.

dguenms:
I think there is still a bug in the stability overlay: when I select core/settlervalue and then change the "paint brush" in the dropdown menu, the overlay disappears.

MeuhV1:
Fixed the overlay disappearing and some other stuff

dguenms:
Okay will look into it again.

dguenms:
You probably need to adjust the export location for files, for instance my partition D is unformatted and I get an exception when trying to export. I'm sure Python can figure that out dynamically, see here for example: http://stackoverflow.com/questions/6227590/finding-the-users-my-documents-path

dguenms:
Another thing: with the latest commit, I removed the ConvertSave script again, the code can handle savegame changes automatically now, so no need to find a place for the button anymore.

MeuhV1:
I restored the convertsave button to it's original state.

I changed the location for exporting the maps to a new map in the mod folder itself. I found an autofinder which finds the install directory of BTS. Then it goes to Mods/RFC Dawn of Civilization/Export. I think this is practical because no additional folder in my games have to be created manually.
(I had to include some example files in those folders, otherwise I couldn't commit the folders)

MeuhV1:
Rewrote the output part. The output can now be copy-pasted directly in the designated files. See the Egypt ones for example. (Don't use them though. I made some random changes in it)

I see that I somehow redid the first 4 commits, so the are executed twice. I can revert it so the 4 double ones are removed and the 4 after that are combined into 1 big commit. (I don't know if I know how to recommit those later one individually) To cleanup the history.

The closing and reopening of the pull request was just me hitting the wrong button BTW. ;)

dguenms:
If you can somehow clean up the commit history, that would be great. Not sure about the best way of going about that though. You can revert to a specific commit ("revert" in the console, the Tortoise interface should use a similar word) so that you are in a state before the changes after it were committed (but the changes themselves are still there in your working copy).

Also, the branch is currently reported as in conflict, looks like a rebase is required as well.

MeuhV1:
Rebased and redid the commits. Combined the last 4 commits into 1 big one and deleted the double ones.

dguenms:
Oh you can even export the cores, that's awesome.

dguenms:
Haven't had the time to test it myself yet: are the sample maps in the export folder created by the export tool? If so, the formatting is excellent already.

MeuhV1:
Yes, they are.

dguenms:
I'm currently running into a bug after recompiling the DLL. Python complains that SCALES_CHAR does not exist even though you correctly exposed in the DLL. Not sure what is wrong here, but I remember having a similar problem once too. So that's what's holding up the merge right now.

Edit: now I see you didn't actually change anything about this. Weird.

dguenms:
Okay, that was a problem with the develop branch, sorry.

Everything looks good and is ready for a merge except for one minor thing: could you change the filenames of the exported files to have the *.txt extension? Would make it easier to open the created files with your preferred text editor later.

MeuhV1:
Done

dguenms:
Great. Will merge when I can compile the DLL.

PR #69, opened 2016-01-27
Type Tag Changes

SgtBears:
This changes the type tags of all units and buildings to match their ingame name. This loads correctly but more errors will probably become evident later.

The DLL hasn't been compiled so you will have to after you merge this.

What do you think about a Text Error thread on civfanatics where people can note errors, submit better pedia texts and submit corrections. I would then be happy to merge them using github.


Assets/Python/Barbs.py
```diff
@@ -70,12 +70,12 @@
 (900, (25, 23), iNative, 'Chan Chan', 2, con.iArcher, 2),		# Chan Chan
 (900, (69, 52), iIndependent, 'Kyiv', 2, con.iLongbowman, 2),		# Kiev
 (990, (49, 56), iCeltia, '&#193;th Cliath', 1, -1, -1),			# Dublin
-(1000, (61, 63), iIndependent2, 'Nidaros', 1, con.iVikingBerserker, 1),	# Trondheim
-(1000, (71, 17), iNative, 'Quelimane', 1, con.iZuluImpi, 1),		# Quelimane
-(1100, (71, 20), iNative, 'Mombasa', 1, con.iZuluImpi, 1),		# Mombasa
+(1000, (61, 63), iIndependent2, 'Nidaros', 1, con.iVikingHuskarl, 1),	# Trondheim
```
dguenms: Isn't it Huscarl in the game?

Assets/Python/Barbs.py
```diff
@@ -70,12 +70,12 @@
 (900, (25, 23), iNative, 'Chan Chan', 2, con.iArcher, 2),		# Chan Chan
 (900, (69, 52), iIndependent, 'Kyiv', 2, con.iLongbowman, 2),		# Kiev
 (990, (49, 56), iCeltia, '&#193;th Cliath', 1, -1, -1),			# Dublin
-(1000, (61, 63), iIndependent2, 'Nidaros', 1, con.iVikingBerserker, 1),	# Trondheim
-(1000, (71, 17), iNative, 'Quelimane', 1, con.iZuluImpi, 1),		# Quelimane
-(1100, (71, 20), iNative, 'Mombasa', 1, con.iZuluImpi, 1),		# Mombasa
+(1000, (61, 63), iIndependent2, 'Nidaros', 1, con.iVikingHuskarl, 1),	# Trondheim
+(1000, (71, 17), iNative, 'Quelimane', 1, con.iBarbarianImpi, 1),		# Quelimane
```
dguenms: Actually I'd prefer to retain the original civs in those tags, instead of Barbarian, even if they are not in the game.

Assets/Python/Consts.py
```diff
@@ -1487,37 +1485,27 @@
...
-iLongbowman, iIndianBambooBowman, iCrossbowman, iChinaChokonu, iChariot, iEgyptWarChariot, iHittiteHuluganni, iCelticCidainh, iHorseArcher, iTibetanKhampa,
-iCarthageNumidianCavalry, iKushanAsvaka, iKnight, iMongolKeshik, iArabiaCamelarcher, iByzantineCataphract, iThaiChangSuek, iSeljukGhulamWarrior,
-iMandeFarari, iCuirassier, iSpanishConquistador, iPolishWingedHussar, iSiouxMountedBrave,  iMoorishCamelGunner, iCavalry, iRussiaCossack, iMexicoRurales,
-iArgentineGrenadierCavalry, iWarElephant, iKhmerBallistaElephant, iCarthaginianWarElephant, iTank, iGermanPanzer, iModernArmor, iGunship, iCatapult, iKoreanHwacha,
-iTrebuchet, iBombard, iMughalSiegeElephant, iCannon, iFrenchHeavyCannon, iMachineGun, iArtillery, iMobileArtillery, iWorkboat, iGalley, iPolynesianWaka, iBireme, iTrireme,
-iCaravel, iPortugalCarrack, iTamilDharani, iGalleon, iNetherlandsOostindievaarder, iPrivateer, iIndonesianOrangLaut, iMoorishCorsair, iFrigate,
-iShipOfTheLine, iIronclad, iTransport,  iDestroyer, iCanadaCorvette, iBattleship, iMissileCruiser, iStealthDestroyer, iSubmarine, iAttackSubmarine, iCarrier, iAirship, iFighter,
+iNumUnits = 158
```
dguenms: Why does the overall number change?

Assets/Python/Consts.py
```diff
@@ -1707,22 +1682,18 @@
 #leaders
 
 iNumLeaders = 114
-(iLeaderBarbarian, iNativeLeader, iIndependentLeader, iAlexander, iAsoka, iAugustus, iBismarck, iBoudica, iBrennus, iCatherine, iCharlemagne, iChurchill,
```
dguenms, line 1710: I prefer the shortened names because they save time typing. The other corrections (where the name actually changed) are welcome though.

Assets/Python/Contrib/Sevopedia/SevoPediaBuilding.py
```diff
@@ -289,7 +289,7 @@ def getBuildingSortedList(self, iBuildingType):
 		numInfos = 0
 		for iBuilding in range(gc.getNumBuildingInfos()):
 			if gc.getBuildingInfo(iBuilding).isGraphicalOnly() or iBuilding >= con.iNumBuildings: continue
-			if iBuilding in [con.iBabylonGarden, con.iZuluIkhanda, con.iChinesePavillion, con.iIncanTerrace, con.iNativeAmericaTotem]: continue
+			if iBuilding in []: continue
```
dguenms: I suppose that's covered by bGraphicalOnly now? If so, just remove the entire line, this is just dead code.

Assets/Python/RiseAndFall.py
```diff
@@ -892,7 +892,7 @@ def checkTurn(self, iGameTurn):
                         utils.makeUnit(iSettler, iCarthage, (58, 39), 1)
                         utils.makeUnit(iArcher, iCarthage, (58, 39), 2)
                         utils.makeUnit(iWorker, iCarthage, (58, 39), 2)
-                        utils.makeUnit(iCarthaginianWarElephant, iCarthage, (58, 39), 2)
+                        utils.makeUnit(iPhoenicianWarElephant, iCarthage, (58, 39), 2)
```
dguenms: Shouldn't it be African War Elephant?

dguenms:
Nothing besides those inline comments. Also it looks like you have to rebase before I can merge.

And the thread is a good idea. It's probably best if you create it so you can maintain the OP. I will endorse it of course.

SgtBears:
I removed unused buildings and units (like the Incan Granary and such) from the xml code. That's where some of the odd changes came from. I can add them back in if you want but they haven't been used in 100s of versions.
I fixed your comments, removed the spy specialist from the civilopedia and fixed the featureimpassable civilopedia text.
You still have to compile the dll.

dguenms:
I prefer to keep the unused XML entries, actually. You never know when they will be needed again, especially with new civs being added.

SgtBears:
But they are easy enough to add back in when you actually add the new civs. I agree about keeping relevant xml entries but are you ever going to add the zulus or native americans or revert the chinese and babylonians back to the pavilion or garden. So I could add the stepwell and terrace again since they are recent removals but there is no point keeping the garden or pavilion which have been obsolete since Rhye days.

dguenms:
It's even less effort to just keep them in. While there is not much gained by removing them from a practical perspective.

SgtBears:
Okay, added them back in. I am 99.99% sure that various dilectus will be incompatible with these changes.

dguenms:
Thanks. Merge upcoming.

Would you be willing to apply the same tag changes for VD as well? It has its own repository [here](https://github.com/dguenms/DoC-Varietas-Delectat).

PR #70, opened 2016-01-31
Platybuilder update

MeuhV1:
Modified the export tool:
Setting peak and water tiles to a default value is done during initializing
Peak and water tiles can't be changed at all
Peak and water tiles are removed from the exceptions list (the won't be set to core during initializing anyway)
Added force export option: If the shift key is down while clicking on the export button, it will export even if there aren't any changes
If altkey + export button, the maps of all civs will be exported
Updated output for reborn civ cores so it can be copied into the dictionaries
Right-clicking on a tile in settlervalue mode and warvalue mode sets the value to 20 and 0 respectively.

Some small bugfixes


dguenms:
Merge incoming. Sorry that it took so long.

PR #71, opened 2016-01-31
Updated cores, Warmaps, Settlermaps

MeuhV1:
I ran all civs through the export script. They are all the same, but "cleaned up".

Cores:
The script made some core rectangles smaller without the loss of any tiles. (Italy for example)
Some civs don't have an entry in the exceptions dictionary, because these tiles were out of the rectangle anyway. (like Babylonia) Or because they are all water or peak tiles.

Settlermaps:
All peak and water tiles have a value of 20

Warmaps:
All peak and water tiles have a value of 0
All core tiles have a value of at least 8


dguenms:
Good to go and merged.

PR #75, opened 2016-02-13
Various fixes

SgtBears:
Requires you to merge the dll.

I removed the doubled up text keys.
Renamed UNIT_NUMIDIAN_NUMIDIAN_CAVALRY to UNIT_NUMIDIAN_CAVALRY.
Various fixes which resulted from comparing Various Delectus to the normal mod.


dguenms:
So you changed the unit tags to match VD? I don't understand the motivation behind this.

Also why the extra external Varietas Delectat text file?

SgtBears:
No, when I was updating Various Delectat I compared all the VD files with the DOC files and discovered a few bugs (nothing important) in the DOC files.

I changed Numidian Cavalry's type name just cause Imp. Knoedal pointed out that it was stupid.

I moved all text tags that were changed in Various Delectat to a text file in external with all the text tags that were changed. That way it is less likely to cause text errors. I can reverse it I just thought it was a good idea.

dguenms:
I mainly just want to understand the changes.

Why is it better to have the texts changed by VD in another file? It seems that this confuses the "all text keys of one type in the same file" principle that you worked on previously.

I also still don't understand why some art def tags needed to be changed? Were those bugs?

SgtBears:
None of these changes do anything for gameplay. I just thought it was a bit neater.

The art def tags I changed back from UNIT_'era'_WORKER to UNIT_WORKER_'era' it makes it easier to search for a particular art def. It was my mistake and I wanted to fix it.

The Middle_east_era and the Asia_era were based off the Medieval_era so I updated the tech cost for those eras just in case it was ever used.

The reason I moved the units to the VD file is that depending on where the VD file went in the text folder there was a slight chance that the VD text tags would get overwritten by a later text file. The 'all text tags of 1 type in 1 file' isn't perfect already so I figured moving 3 text tags out wouldn't matter. Shall I remove it?

The reason I changed Numidian Cavalry is that Imp Knoedal pointed out that it was stupid calling it NUMIDIAN_NUMIDIAN_CAVALRY even though that technically was correct by my naming standard.

dguenms:
Oh okay, I thought the art tag changes were also related to VD. That's alright then.

And in fact, yeah, I would prefer if you reverted the external file for VD texts.

Everything else is fine, thanks for the explanations.

SgtBears:
Fixed

dguenms:
Great. Will merge when I have a compiler on hand to push the new DLL.

PR #76, opened 2016-02-14
Wondermovie updates

MeuhV1:
Fixed double wondermovies for projects
Wondermovies only appear the first time a buildings/wonder is build
Removed condition that movie only appear for world wonders

### DIFF
--- a/Assets/Python/CvEventManager.py
+++ b/Assets/Python/CvEventManager.py
@@ -543,7 +543,7 @@ class CvEventManager(object):
 		'Building Completed'
 		pCity, iBuildingType = argsList
 		game = gc.getGame()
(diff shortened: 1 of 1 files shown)


dguenms:
What are the double project movies about? Are the movies also triggered by the DLL so they appear twice when the Python code is also executed?

MeuhV1:
I don't think the DLL has anything to do with it. The python code that triggers it was coded twice, so I removed one trigger. This is probably introduced when merging platybuilder. Wonders did have the same problem, but I fixed that already in a previous pull request.

dguenms:
Oh it's from the Platybuilder merge, got it. Merged.

PR #77, opened 2016-02-14
Resources and wonder movies

Imperator-Knoedel:
-Added Sugar in Suriname and some strategic resources in India, China and Australia as well as a forest in the Netherlands.

-Added wonder movies for most shrines, SDI, Apollo Program, national wonders (doesn't seem to work though), academies (doesn't seem to work though) and military academies (doesn't seem to work though) and some world wonders.


dguenms:
Looks good, I would merge it as is.

However, you mentioned that you still want to remove the now unused still images. As I said you can include that here by simply pushing to the branch belonging to this pull request. I'll be waiting for that.

Imperator-Knoedel:
I... think I just did that? Maybe? There should be two files missing, Brandenburg and ForbiddenPalace or something, the rest was either overwritten because same name or in the case of the shrines is a Civ4 Vanilla file and thus doesn't need to be (or is actually able to) deleted.

dguenms:
You did. Merged.

PR #78, opened 2016-02-21
Python cleanup

MeuhV1:
Aligned tabs in python files (replaced 8 spaces with a tab) This gives a better display of the code in Notepad++

Included new civic screen with combo info

### DIFF
--- a/Assets/Python/AIWars.py
+++ b/Assets/Python/AIWars.py
@@ -2,7 +2,7 @@
 
 from CvPythonExtensions import *
 import CvUtil
-import PyHelpers        # LOQ
+import PyHelpers	# LOQ
 import Popup
 #import cPickle as pickle
 import Consts as con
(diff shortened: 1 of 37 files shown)


dguenms:
That was long overdue, thanks for taking care of that. Will merge soon.

dguenms:
Sorry, another merge caused a conflict. Hard to avoid when almost all files are affected :/

MeuhV1:
That's true. But I updated my branch so the conflicts are gone.

MeuhV1:
I've also included some changes in the civic screen to show good/bad combos
(I think it's unnecessary to create another branch for this)

dguenms:
Okay, but I will have to look at it in game.

Thanks for making the branch compatible again.

dguenms:
Okay, looks good. The code mght need to be refactored again later so that Stability.py and the display rely on the same definitions but that's not required right away.

PR #79, opened 2016-02-24
Update civic combination stability calculations

MeuhV1:
Update civic combination stability
Pros:
Civic info advisor is automatically synchronized
Actual calculation code much more compact

Cons:
Maintaining the combination matrix is probably a bit more work

### DIFF
--- a/Assets/Python/Screens/CvCivicsScreen.py
+++ b/Assets/Python/Screens/CvCivicsScreen.py
@@ -3,6 +3,7 @@ import CvUtil
 import CvScreenEnums
 import CvScreensInterface
 import Consts as con
+import Stability
 
 # Globals
 gc = CyGlobalContext()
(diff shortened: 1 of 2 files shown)


dguenms:
With a matrix this sparse, there should be a better way of maintaining that data.

MeuhV1:
True. While I was making it, I was thinking of abandoning this update for that reason. But as it wasn't that much work, I decided to finish it so you can give your opinion. (I mean, it does work, albeit with some drawbacks) I did think of another method which is very similar to the current method. I will recommit it when it's done (overwriting this commit), so you can take a look at it.

MeuhV1:
I committed the new method of getting civic combo values. I created a new function which return the value for a combination of 2 civics. It works very similar to the old method.

dguenms:
I think this is still somewhat flawed because it expects civics to be inserted in a specific order, but I already have an idea on how to improve that, so I'm just going to merge it. Thanks for the interface work!

MeuhV1:
No, the order doesn't matter. The for loop takes care of that.

dguenms:
You're right, I had already wondered why you did that but then misinterpreted the code.

PR #81, opened 2016-03-18
Fixed core change Phoenicia

MeuhV1:


dguenms:
Thanks. What's with the removed line in isForeignCore()?

MeuhV1:
That line is a leftover from the old isForeignCore function. I loop through civs. It makes no sense if a plot itself continues that loop.

dguenms:
Oh right, I see. Code merged.

PR #86, opened 2016-04-07
Update DynamicNames_Babylonia.xml

citisDoC:
So that the "Assyrian Empire" displays properly, if you build your capital in Ninua.


dguenms:
Good catch, and thanks for the quick fix!

PR #93, opened 2016-04-09
Satrapy Fix and URV Fix

SgtBears:


dguenms:
That's helpful, thanks.

PR #98, opened 2016-04-20
Fixes

SgtBears:
Reverted MissionInfos


dguenms:
Thanks for the quick fix.

PR #102, opened 2016-04-24
MERGE SECOND - Standardizes Civilization names in the DLL

SgtBears:
Also changes ottoman to Turkish in the changelog. I am changing other tags later hence these changes.


dguenms:
Good to go, and I think this can still be merged before the first one.

PR #106, opened 2016-04-25
Might help with Hotkeys

SgtBears:
Adds back in the version =1.0 line


dguenms:
I don't think that's the issue actually, and it works fine for me.

PR #111, opened 2016-04-27
Pedia fixes + updates

MeuhV1:
Fixed missing promotions in units page
Fixed error when trying to open Coal and Oil resource page

Added Ctrl-K (stability overlay) info in shortcuts
Slightly adjusted some info text panels (replaced some unnecessarily large gaps with smaller ones, so it can display more text)
Made some text auto translate
Added religious victory conditions to religions page + pantheon/secularism page
Removed some code that is HR specific (like Natural Wonders, Fake tech)


dguenms:
I don't really like the appearance conditions being detailed in the Civilopedia, that is something players should discover for themselves while playing the game imo. Plus as you mention I don't want to create another obligation to update the documentation in case I want to change something.

With the text panels do you mean the bottom text panel and its distance to the bottom of the screen? Or other distances as well?

MeuhV1:
Ok, I'll remove the appearing conditions.

I mean this gap. It was too big and gave me a feeling something wasn't right.
http://forums.civfanatics.com/showpost.php?p=14249394&postcount=1254

dguenms:
Oh, the text box and the panel were not aligned, yeah that does look odd. Notify me when you've updated the PR.

MeuhV1:
Done. I overwrote the last commit, so the log keeps clean.

dguenms:
Thanks.

I think the civic/religion screens look a bit too crowded with the extra victory panel. Can you instead put the goals at the top of the history text like it was in the old civilopedia?

MeuhV1:
No problem. Should I also do that for the religions page?

dguenms:
Yeah, for both of the pages.

MeuhV1:
Done

dguenms:
Great, and merged.

PR #115, opened 2016-05-04
Diplomacy Bug Fix

SgtBears:
Removed the commented out Ivan 4 diplomacy. This fixes the current bug that some first contact diplomacy texts didn't appear. I don't know how, it just does.

Also adds all the other leaders who didn't have a first contact to the generic contact.

I also uploaded an updated Soundtrack Module.


dguenms:
Thanks, I noticed recently that many diplo texts were missing where I thought we had some defined. Weird that comments cause that issue but I read about something similar recently.

PR #119, opened 2016-05-11
Python to Scenario File

SgtBears:
I moved gold, starting units, 1700 added culture, great specialists, diplomacy bonuses and the 600 AD byzantine flip from python to the scenario file.

The 600 AD scenario was too big to display by itself in a seperate pull request and I assume that the 1700 AD scenario is the same. I can provide the files on CFC if you want.

The reason why I didn't do this in stages is that I saved the scenario file when all the python was loaded and then just removed all the unwanted information (like technologies) from there.

### DIFF
--- a/Assets/Python/RiseAndFall.py
+++ b/Assets/Python/RiseAndFall.py
@@ -457,12 +457,12 @@ class RiseAndFall:
 		if utils.getScenario() == i600AD:
 			Civilizations.initScenarioTechs(i600AD)
 			self.create600ADstartingUnits()
-			self.assign600ADGold()
+#			self.assign600ADGold()
 			
 		if utils.getScenario() == i1700AD:
 			Civilizations.initScenarioTechs(i1700AD)
 			self.create1700ADstartingUnits()
-			self.assign1700ADGold()
+#			self.assign1700ADGold()
 			self.init1700ADDiplomacy()
 			self.prepareColonists()
 			self.adjust1700ADCulture()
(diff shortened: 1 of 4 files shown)


Assets/Python/RiseAndFall.py
```diff
@@ -3542,119 +3542,120 @@ def create1700ADstartingUnits(self):
 	
 		# China
 		tCapital = tBeijing
-		utils.makeUnit(iMusketman, iChina, tCapital, 12)
```
dguenms: Where did you move the starting units? They don't seem to appear in the scenario file.

dguenms:
If it's all still there then alright.

Assets/Python/RiseAndFall.py
```diff
@@ -538,23 +538,23 @@ def adjust1700ADCulture(self):
 				if plot.getOwner() != -1:
 					utils.convertPlotCulture(plot, plot.getOwner(), 100, True)
 					
-		for x, y in [(48, 45), (50, 44), (50, 43), (50, 42), (49, 40)]:
```
dguenms, line 541: Does this work from the scenario file? I.e. is there no other culture left on those tiles so they won't flip shortly after?

dguenms:
Also, please make sure to remove all the commented lines before I can merge this.

SgtBears:
All I though that code did was add 100 of the noted player's culture to the tile. So that was what I did. I can revert that, but adding 100 culture will probably be enough by itself.

SgtBears:
I merged the pull requests into 1 and removed all the commented lines.

I also noticed that Byzantiums cultural borders are slightly different than before, but nothing important.

dguenms:
The culture code converts 100% culture of other players to the owner. So please revert.

SgtBears:
Reverted

PR #120, opened 2016-05-11
Python to Scenario Part 2

SgtBears:
The 1700 AD scenario to view it all requires splitting it into 4 or 5 pieces so It would be easier for me to upload it to CFC. Most of the changes are very simple and easy to see why they were made. There is just a lot of them.


dguenms:
No, it's fine. Waiting for the preceding pull request.

PR #124, opened 2016-05-15
Makes the BUG building actual effects option the default bug setting

SgtBears:
I have a feeling that most players use this or would if they knew about it. I could do a poll on civ fanatics if you want to determine if it should be the default option.

The Files changed section is displaying 100% wrong I don't know why. I just changed all the Building Actual Effects options to true.


dguenms:
Why did you close this?

PR #131, opened 2016-06-26
Python updates and relevant fixes for new HR tech tree port

temujin2k:
Hello, Leoreth!

I took some time to update the Python files in accordance with the new tech tree update. I have already finished adapting the spawn techs for every civ in all three scenarios to the new tech tree, but the lists I wrote there are based on my intuition and rough estimations of the equivalent spawn tech lists for the original tech tree. I also substituted many of the tech constants in the affected files. However I'm not so sure if I made the correct substitutions for the techs.

I've already tested a short game for each scenario with the new changes I made, and so far the game loads without errors, although minor Python exceptions may pop up along the way. I found that teching is still very slow here, so I think the DLL needs a lot of updating too. But I don't have any knowledge on C++ or editing the core DLL files so I won't be able to touch that.

_PS: I deleted, updated and pushed a new branch and then made a new pull request because I managed to screw up my commit history a few hours ago and the AppVeyor Bot checks that were too many messed up the pull request page. I'm very sorry about that!_


dguenms:
Hey, thanks for the pull request. No need to worry about the builder bot, I don't think it's completely perfectly configured right now anyway.

Right now I'm not really in the state to be able to accept pull requests for the ongoing work though. Everything is still in a very fluid and experimental to be dealing with stuff like Python code, which is why I haven't touched it yet. And when I do, I really want to have a lot of conscious control over the changes I am making. Will return to your pull request for reference though when I get there.

temujin2k:
It's fine! I just update what needs to be updated whenever you push new commits. I guess changes you'll make to the tech tree will not need major Python revisions to this one, especially since the order of the techs in the XML made things much more convenient. Can't wait for the final tech tree and the relevant changes! I'll just keep this PR open just in case. :)

PR #133, opened 2016-08-06
New options

MeuhV1:
- Added some new options: No Human Stability, No stability all civs, unlimited switching (All off by default)
- All options can be changed during the game in the WB (gameoptions menu)
- Added function to enable/disable secondary civs in the gameoption menu
- Moved setPlayerEnabled function to StoredData.py (It was in several places. Now it is in just one place)

Several people asked how to to modify their game to incorporate these changes. These new options make it a lot easier for them.

### DIFF
--- a/Assets/Python/StoredData.py
+++ b/Assets/Python/StoredData.py
@@ -37,6 +37,7 @@ class StoredData:
 				    'lTempPlots': [],
 				    'iSpawnWar': 0, #if 1, add units and declare war. If >=2, do nothing
 				    'bAlreadySwitched': False,
+					'bUnlimitedSwitching': False,
 				    'lColonistsAlreadyGiven': [0 for i in range(con.iNumPlayers)], #active players
 				    'lAstronomyTurn': [1500 for i in range(con.iNumPlayers)], #active players
 				    'lNumCities': [0 for i in range(con.iNumTotalPlayers)], #total players to contain Byzantium too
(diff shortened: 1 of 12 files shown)


dguenms:
Good idea.

The only point I would make is to rename the two no stability options to no collapse, because the stability mechanic works regardless of the setting as far as I can tell. Alternatively, you can also abort all stability calculations with the corresponding options set - this should also save some computational power.

MeuhV1:
I'm aware that the calcuations still run in the background. I did consider aborting calculating all stability calcuations, but I did the easy, lazy option. (And I didn't want to miss a trigger, because I was afraid the would cause errors)

But now I see the easy way to disable the calculations. Just add a return statement to all functions triggered by other files which do stability calculations in stability.py. That should do it, right?

dguenms:
The way I see it the central component of Stability.py (and also the most computationally expensive) is calculateStability(). Just adding a "return 0" right at the beginning if the option is selected should save a lot of time and not break anything. Leave the current returns for the collapse checks in place to be sure, but otherwise I don't think you have to mess with the triggers, their contribution to the calculation time is probably minimal.

MeuhV1:
Updated the PR.

I thought of adding a new option which allows to change the stability level of each civ in the WB. I decided not to add it, because your stability level will return to the old situation after a couple of checks. So it doesn't have a significant impact.

Assets/Python/StoredData.py
```diff
@@ -150,6 +153,24 @@ def getTimedConquests(self):
 	def resetTimedConquests(self):
 		self.scriptDict['lTimedConquests'] = []
 		
+	def setPlayerEnabled(self, iCiv, bNewValue):
+		self.scriptDict['lPlayerEnabled'][con.lSecondaryCivs.index(iCiv)] = bNewValue
+		
+	def getPlayerEnabled(self, iCiv):
```
dguenms: Just a small comment, by convention getter methods returning a boolean are usually called isPlayerEnabled() etc., same for the other ones here.

dguenms:
Looks good!

MeuhV1:
Updated

dguenms:
Great, merged.

PR #134, opened 2016-08-12
Bug fixes in congresses and Platybuilder updates

MeuhV1:
Fixed bug that causes congresses not to happen if countdown is lowered with WB
Fixed bug when the game tries to host the congress in a city of a non living civ
Added possibility to increase the congress countdown turns

Added some "HideInactive" buttons
Updated the modifiers editor so it uses adjusted values when resetting in late start scenarios


MeuhV1:
EDIT: Updated the PR upto and including the StoredData-to-class update.

PR #135, opened 2016-08-14
Code optimization and cleanup

MeuhV1:
Various code optimization in Plague and RFCUtils
Prevented 1650 plagues starting in 1700 AD scenario

Code Cleanup:
Changed various pCurrent into pPlot (This makes it more obvious that it deals with plots)
Removed isNoVassal() function (just use not isAVassal() function)
Removed tabs and spaces at the end of lines, replaces spaces with tabs for consistency

### DIFF
--- a/Assets/Python/Plague.py
+++ b/Assets/Python/Plague.py
@@ -3,13 +3,13 @@
 
 from CvPythonExtensions import *
 import CvUtil
-import PyHelpers   
+import PyHelpers
 #import Popup
 #import cPickle as pickle
 from Consts import *
 from StoredData import data #edead
-import RFCUtils
-utils = RFCUtils.RFCUtils()
+from RFCUtils import utils
+import random
 
 # globals
 gc = CyGlobalContext()
@@ -19,22 +19,25 @@ iDuration = 6
 
 class Plague:
 
-#######################################
+######################################
 ### Main methods (Event-Triggered) ###
-#####################################  
+######################################
 
 
 	def setup(self):
 		for i in range(iNumMajorPlayers):
 			data.players[i].iPlagueCountdown = -utils.getTurns(iImmunity)
 		
-		if utils.getScenario() == i3000BC:  #late start condition
+		if utils.getScenario() == i3000BC:  #early start condition
 			data.lGenericPlagueDates[0] = getTurnForYear(400) + gc.getGame().getSorenRandNum(40, 'Variation') - 20
 		else:
 			data.lGenericPlagueDates[0] = 80
 			
 		data.lGenericPlagueDates[1] = getTurnForYear(1300) + gc.getGame().getSorenRandNum(40, 'Variation') - 20
-		data.lGenericPlagueDates[2] = getTurnForYear(1650) + gc.getGame().getSorenRandNum(40, 'Variation') - 20
+		if utils.getScenario != i1700AD:
+			data.lGenericPlagueDates[2] = getTurnForYear(1650) + gc.getGame().getSorenRandNum(40, 'Variation') - 20
+		else:
+			data.lGenericPlagueDates[2] = 300 # Safe value to prevent plague at start of 1700 scenario
 		data.lGenericPlagueDates[3] = getTurnForYear(1850) + gc.getGame().getSorenRandNum(40, 'Variation') - 20
 
 		undoPlague = gc.getGame().getSorenRandNum(8, 'undo')
@@ -52,7 +55,7 @@ class Plague:
 					data.players[i].iPlagueCountdown -= 1
 					if (data.players[i].iPlagueCountdown == 2):
 						self.preStopPlague(i)
-					if (data.players[i].iPlagueCountdown == 0):
+					elif (data.players[i].iPlagueCountdown == 0):
 						self.stopPlague(i)
 				elif (data.players[i].iPlagueCountdown < 0):
 					data.players[i].iPlagueCountdown += 1
@@ -95,24 +98,21 @@ class Plague:
 
 		iWorstCiv = -1
 		iWorstHealth = 200
-		for i in range(iNumMajorPlayers):
-			pPlayer = gc.getPlayer(i)
+		for iPlayer in range(iNumMajorPlayers):
+			pPlayer = gc.getPlayer(iPlayer)
 			if (pPlayer.isAlive()):
-				if (self.isVulnerable(i) == True):
-					iHealth = -30
-					iHealth2 = iHealth/2
+				if self.isVulnerable(iPlayer):
+					iHealth = self.calculateHealth(iPlayer)
+					iHealth /= 2
 					if (pPlayer.calculateTotalCityHealthiness() > 0):
-						iHealth = int((1.0 * pPlayer.calculateTotalCityHealthiness()) / (pPlayer.calculateTotalCityHealthiness() + \
-							pPlayer.calculateTotalCityUnhealthiness()) * 100) - 60
-						iHealth2 = iHealth/2
-						iHealth2 += gc.getGame().getSorenRandNum(40, 'random modifier')
-						#print ("starting plague", "civ:", i, "iHealth:", iHealth, iHealth2)
+						iHealth += gc.getGame().getSorenRandNum(40, 'random modifier')
+						#print ("starting plague", "civ:", iPlayer, "iHealth:", iHealth, iHealth2)
 						if (iPlagueCounter == 2): #medieval black death
-							if (i in lCivGroups[0]):
-							       iHealth2 -= 5 
+							if (iPlayer in lCivGroups[0]):
+								iHealth2 -= 5 
 					if (iHealth2 < iWorstHealth):
 						iWorstHealth = iHealth2
-						iWorstCiv = i
+						iWorstCiv = iPlayer
 		if (iWorstCiv != -1):
 			print ("worstCiv", iWorstCiv)
 			pWorstCiv = gc.getPlayer(iWorstCiv)
@@ -127,12 +127,7 @@ class Plague:
 
 	def preStopPlague(self, iPlayer):
 
-		cityList = []
-		apCityList = PyPlayer(iPlayer).getCityList()
-		for pCity in apCityList:
-			city = pCity.GetCy()
-			if (city.hasBuilding(iPlague)):
-				cityList.append(city)
+		cityList = [city for city in utils.getCityList(iPlayer) if city.hasBuilding(iPlague)]
 		if (len(cityList)):
 			iModifier = 0
 			for city in cityList:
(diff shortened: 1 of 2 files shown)


MeuhV1:
EDIT:
Updated upto and and including the StoredData update.
Didn't include the code cleanup (Removal of tabs at end of lines and removal of brackets).

MeuhV1:
Updated to latest version.

Assets/Python/RFCUtils.py
```diff
@@ -998,25 +968,13 @@ def getTurns(self, turns): # edead
...
-		gc.getMap().plot(2, 2).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(1, 2).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(0, 2).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(122, 0).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(122, 1).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(122, 2).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(123, 2).setRevealed(iCiv, False, True, -1);
+			catapult = plotZero.getUnit(0)
+			catapult.kill(False, iCiv)
+		for x in range(-2,3):
```
dguenms: There also is a surroundingPlots() method in RFCUtils.py now, which should be used for cases like this. It doesn't use module arithmetic yet (i.e. it would cut off on the map edges for the x axis), but that could be fixed/extended.

Assets/Python/RFCUtils.py
```diff
@@ -550,31 +516,31 @@ def convertPlotCulture(self, plot, iPlayer, iPercent, bOwner):
...
 		#	city.setCulture(iCiv, iCivCulture + iLoopCivCulture, True)  
 	
 ##		for iLoopCiv in range(iNumTotalPlayers):
 ##			if (iLoopCiv != iCiv):
-##				iLoopCivCulture = pCurrent.getCulture(iLoopCiv)
-##				iCivCulture = pCurrent.getCulture(iCiv)
-##				pCurrent.setCulture(iLoopCiv, iLoopCivCulture*(100-iPercent)/100, True)
-##				pCurrent.setCulture(iCiv, iCivCulture + iLoopCivCulture*iPercent/100, True)
-		#iCivCulture = pCurrent.getCulture(iCiv)
+##				iLoopCivCulture = plot.getCulture(iLoopCiv)
```
dguenms, line 529: I know this code is commented out, but why is it inconsistently `plot` and `pPlot`?

Assets/Python/RFCUtils.py
```diff
@@ -382,11 +350,9 @@ def flipCity(self, tCityPlot, bFlipType, bKillUnits, iNewOwner, iOldOwners):
...
 							unit.kill(False, iNewOwner)
 							
-					if (bFlipType): #conquest
-						if (city.getPopulation() == 2):
-							city.setPopulation(3)
-						if (city.getPopulation() == 1):
-							city.setPopulation(2)
+					if bFlipType: #conquest
+						if city.getPopulation() <= 2:
+							city.setPopulation(city.getPopulation() + 1)
```
dguenms: We also have CvCity::changePopulation(iChange).

Assets/Python/RFCUtils.py
```diff
@@ -340,23 +321,10 @@ def flipUnitsInArea(self, lPlots, iNewOwner, iOldOwner, bSkipPlotCity, bKillSett
...
-							gc.getMap().plot(2, 65).setRevealed(iCiv, False, True, -1);
-							gc.getMap().plot(1, 65).setRevealed(iCiv, False, True, -1);
-							gc.getMap().plot(0, 65).setRevealed(iCiv, False, True, -1);
-							gc.getMap().plot(122, 67).setRevealed(iCiv, False, True, -1);
-							gc.getMap().plot(122, 66).setRevealed(iCiv, False, True, -1);
-							gc.getMap().plot(122, 65).setRevealed(iCiv, False, True, -1);
-							gc.getMap().plot(123, 65).setRevealed(iCiv, False, True, -1);
-			
+						if not bRevealedZero:
+							for x in range(-2,3):
```
dguenms: Same here, would be cool to use utils.surroundingPlots() here and extend it to wrap around the x axis.

Assets/Python/RFCUtils.py
```diff
@@ -255,26 +248,14 @@ def flipUnitsInCityAfter(self, tCityPlot, iCiv):
...
-		gc.getMap().plot(2, 67).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(2, 66).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(2, 65).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(1, 65).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(0, 65).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(122, 67).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(122, 66).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(122, 65).setRevealed(iCiv, False, True, -1);
-		gc.getMap().plot(123, 65).setRevealed(iCiv, False, True, -1);
+		for x in range(-2,3):
```
dguenms: See below.

Assets/Python/RFCUtils.py
```diff
@@ -46,14 +46,11 @@ def getGoalsColor(self, iPlayer): #by CyberChrist
...
+
 	#Plague
 	def getRandomCity(self, iPlayer):
-		cityList = []
-		apCityList = PyPlayer(iPlayer).getCityList()
-		for pCity in apCityList:
-			cityList.append(pCity.GetCy())
-		if (len(cityList)):	   
+		citylist = self.getCityList(iPlayer)
+		if len(cityList):
```
dguenms: To check if a list is empty, you can also simply do `if cityList:`.

Assets/Python/Plague.py
```diff
@@ -411,51 +378,46 @@ def isVulnerable(self, iPlayer):
...
 	def onCityRazed(self, city, iNewOwner):
 		if (city.hasBuilding(iPlague)):
 			if data.players[iNewOwner].iPlagueCountdown > 0:
-				apCityList = PyPlayer(iNewOwner).getCityList()
 				iNumCitiesInfected = 0
-				for pCity in apCityList:
-					otherCity = pCity.GetCy()
-					if (otherCity.getX() != city.getX() or otherCity.getY() != city.getY()): #because the city razed still has the plague
+				for otherCity in utils.getCityList(iNewOwner):
+					if city.getID() != otherCity.getID(): #because the city razed still has the plague
```
dguenms: I think Rhye is checking for a new city on the same plot here. When a city is conquered or changes owners in another way, it has a different ID afterwards iirc.

Assets/Python/Plague.py
```diff
@@ -243,40 +228,34 @@ def processPlague(self, iPlayer):
...
-									if (city1.isConnectedTo(city2)):
-										##print ("infect distant", city1.getName(), "to", city2.getName(), utils.calculateDistance(city1.getX(), city1.getY(), city2.getX(), city2.getY()))
-										if (utils.calculateDistance(city1.getX(), city1.getY(), city2.getX(), city2.getY()) <= 6):
-											#print ("infect distant", city2.getName(), "to", city1.getName())
-											self.infectCity(city1)
-											return
-
-
-	def killUnitsByPlague(self, city, pCurrent, baseValue, iDamage, iPreserveDefenders):
+		if len(lNotInfectedCities):
```
dguenms: `if lNotInfectedCities` also works here.

Assets/Python/Plague.py
```diff
@@ -243,40 +228,34 @@ def processPlague(self, iPlayer):
...
+			if sd.getPlagueCountdown(iPlayer) > 2: #don't spread the last turns
```
dguenms: Both sd and getPlagueCountdown() shouldn't exist anymore.

Assets/Python/Plague.py
```diff
@@ -243,40 +228,34 @@ def processPlague(self, iPlayer):
...
+										#print ("infect by trade route", city.getName(), "to", loopCity.getName())
+									elif gc.getTeam(pPlayer.getTeam()).isOpenBorders(iOwner) or \
+										gc.getTeam(pPlayer.getTeam()).isVassal(iOwner) or \
+										gc.getTeam(gc.getPlayer(iOwner).getTeam()).isVassal(iPlayer): #own city, or open borders, or vassal
+										if (self.isVulnerable(iOwner) == True):
+											self.spreadPlague(iOwner)
 											self.infectCity(loopCity)
 											#print ("infect by trade route", city.getName(), "to", loopCity.getName())
+											iHuman = utils.getHumanID()
+											if (gc.getPlayer(iHuman).canContact(iOwner) and iHuman != iOwner):
```
dguenms: Maybe add an `loopCity.plot().isRevealed(iHuman)` here too? Would avoid the interface clutter with undiscovered cities.

Assets/Python/Plague.py
```diff
@@ -170,68 +164,59 @@ def processPlague(self, iPlayer):
...
-									if (self.isVulnerable(iLoopCiv) == True):
-										if data.players[iPlayer].iPlagueCountdown > 2: #don't spread the last turns
+					if data.players[iPlayer].iPlagueCountdown > 2: #don't spread the last turns
+						if not data.players[iPlayer].bFirstContactPlague: #don't infect if first contact plague
+							for iLoopCiv in range(iNumMajorPlayers):
+								if (gc.getTeam(pPlayer.getTeam()).isVassal(iLoopCiv) or \
+									gc.getTeam(gc.getPlayer(iLoopCiv).getTeam()).isVassal(iPlayer)):
+									if (gc.getPlayer(iLoopCiv).getNumCities() > 0): #this check is needed, otherwise game crashes
+										capital = gc.getPlayer(iLoopCiv).getCapitalCity()
+										if (self.isVulnerable(iLoopCiv) == True):
```
dguenms: `self.isVulnerable(iLoopCiv)` is fine, no need to check for boolean equality.

dguenms (changes requested):
Great refactoring, some minor stuff could be improved even further, would be nice if you could include that as well.

Assets/Python/RFCUtils.py
MeuhV1: That is because argument of the function this code is part of named it plot. Everything in the function itself is called plot, but in the commented section it was still called pCurrent. I only synchronized the commented section with the rest of the function.

MeuhV1:
Updated the PR with your suggestions.

I didn't touch the surroundingPlots function. For unrevealing the plots, the function should return all tiles within the selected radius from the plot. But now, the function only returns plots that are the radius away from the input plot. (Ergo, if the radius is 2, it returns the plots in the second ring of the the input plot, but not the first ring)

Assets/Python/RFCUtils.py
dguenms: Okay, that makes sense.

dguenms:
Okay, seems like I need to take another look at the surroundingPlots() function. Ready to merge otherwise.

dguenms (approved):

PR #137, opened 2016-11-03
StoredData editor in WB and StabilityMap editor overhaul

MeuhV1:
StoredData editor in WB
Couple of code updates

The editor is able to change integers, boolean and lists containing lists and booleans. If you add a new variable to the StoredData, the editor will automatically be able to handle them. (As long as the variable are of the mentioned types)

Stability maps editor overhaul:
- All map editors are now in a new mode, instead of being part of the map mode
- 2 new map editors/viewers: Religion map editor and Region map editor
- New option to select multiple tiles: Region mode; selects all tiles in a region (Also works in the map mode)
- Various code improvements/cleanup


MeuhV1:
The current version has a small bug in it. I'm working on an update which includes a fix for this. Don't merge until that is fixed.

MeuhV1:
Updated to latest version.

Assets/Python/DynamicCivs.py
```diff
@@ -1940,7 +1940,7 @@ def leader(iPlayer):
 	elif iPlayer == iSpain:
 		if isFascist(iPlayer): return iFranco
 		
-		if 1 in data.lFirstContactConquerors: return iPhilip
+		if True in data.lFirstContactConquerors: return iPhilip
```
dguenms, line 1943: Good catch.

Assets/Python/Screens/PlatyBuilder/WBPlotScreen.py
```diff
@@ -619,13 +623,13 @@ def handleInput(self, inputClass):
 		elif inputClass.getFunctionName().find("BaseYield") > -1:
 			i = YieldTypes(inputClass.getData2())
 			if inputClass.getData1() == 1030:
-				CyGame().setPlotExtraYield(pPlot.getX(), pPlot.getY(), i, iChange)
+				CyGame().setPlotExtraYield(pPlot.getX(), pPlot.getY(), i, abs(iChange))
```
dguenms, line 626: Can you explain what's up with the absolute values here? Shouldn't a negative change be possible?

dguenms (commented):
The code looks okay from a superficial glance. Will test it in the game later.

Assets/Python/Screens/PlatyBuilder/WBPlotScreen.py
MeuhV1: I changed the mechanics of the global variable iChange. Previously, iChange was just a positive integer. There was a different boolean which made that value positive or negative. Many screen had a dropdown menu which switches that boolean between True and False.

When you change something, like the yield production of a tile, it first looks at the boolean value to see if it should add or remove and then change it by the value iChange.
The example you highlighted is a case when there are 2 buttons to change the value. The + button should always increase the value and the - button should always decrease the value.

I changed it so that the add/remove dropdown menu doesn't change that specific boolean, but it switches the sign of the integer. (So iChange can be positive or negative, depending on the mode)
To make sure that the buttons still work properly, I had to make it absolute values. They make sure a + buttons always add and the - buttons always subtract, no matter if iChange is positive or negative.

The reason why I did this is because the adding and removing consequent in all WB screen files. I synced all files to this respect. And it makes some changing function simpler because it only has to change the value by iChange. (It will automatically increase or decrease depending on the mode you have selected)

dguenms: Okay, that makes sense.

PR #140, opened 2017-01-16
Improved Great People

etiennefd:
At first I was going to simply correct a few mistakes here and there, and maybe add some names I could think of. I ended up adding quite a lot. 

Highlights: 
- fixed many spelling mistakes, missing diacritics, missing first names, missing feminine markers, wrong types of GP, wrong eras, and names that were not in their native language
- the German merchant and engineer lists were identical; I split them and added several names
- added some non-prophet names for Tibet, then realized Tibet uses Chinese lists, and figured I'd leave the names in anyway
- tried to improve ancient or poorly known civilization lists (Egypt, Babylon, Phoenicia, Aztecs), which often include random ruler names that have nothing to do with their type. It's better now, especially for artists and scientists, though it's still not perfect.

~ Steb

### DIFF
--- a/Assets/Python/GreatPeople.py
+++ b/Assets/Python/GreatPeople.py
@@ -1199,6 +1199,7 @@ iCivFrance : {
 		u"Marcel Légaut", # 20th
 	],
 	iGreatArtist : [
+		u"Pérotin", # 12th
 		u"Chrétien de Troyes", # 12th
 		"fChristine de Pizan", # 15th
 		iRenaissance,
@@ -1333,6 +1334,8 @@ iCivEngland : {
 		"Francis Drake", # 16th
 		"James Cook", # 18th
 		"Adam Smith", # 18th
+		iIndustrial,
+		"Cecil Rhodes", # 19th
 		iModern,
 		"John Maynard Keynes", # 20th
 	],
--- a/Assets/Python/GreatPeople.py
+++ b/Assets/Python/GreatPeople.py
@@ -456,6 +456,7 @@ iCivGreece : {
 iCivCarthage : {
 	iGreatProphet : [
 		"Acherbas", # legendary
+		"fJezebel", # 9th BC
 		iClassical,
 		"Tertullian", # 2nd
 		"Cyprian", # 3rd
@@ -474,6 +475,7 @@ iCivCarthage : {
 		"Bomilcar", # 3rd BC
 		"Eshmuniaton", # ?
 		"Tanit", # ?
+		"Sanchuniathon", # unknown date
 	],
 	iGreatMerchant : [
 		"Hanno", # 5th BC
@@ -575,7 +577,6 @@ iCivRome : {
 	],
 	iGreatScientist : [
 		"Cato", # 1st BC
-		"Cicero", # 1st BC
 		"Sosigenes", # 1st BC
 		"Plinius", # 1st
 		"Strabo", # 1st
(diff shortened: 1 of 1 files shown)


Assets/Python/GreatPeople.py
```diff
@@ -486,18 +513,21 @@ def setup():
...
 	],
 	iGreatStatesman : [
-		"Astarymus", # 9th BC
+		"Ashtar-rom", # 9th BC
+		"Ithobaal", # 9th BC
 		"Pygmalion", # 9th BC
 		"fElishat", # 8th BC
-		"Hanno", # 5th BC
-		"Azelmicus", # 4th BC
+		"Hanno", # 4th BC
```
dguenms, line 522: Maybe we should add a split into the classical era with the 4th century BC onwards?

Assets/Python/GreatPeople.py
```diff
@@ -629,11 +660,14 @@ def setup():
 		"Saigyo Hoshi", # 12th
 		iRenaissance,
 		"Kano Eitoku", # 16th
-		iIndustrial,
 		"Toshusai Sharaku", # 18th
-		"Katsushika Hokusai", # 18th
+		"Yosa Buson", # 18th
+		iIndustrial,
+		"Katsushika Hokusai", # 19th
```
dguenms, line 668: 👍

Assets/Python/GreatPeople.py
```diff
@@ -666,15 +700,15 @@ def setup():
 		"Shigeru Miyamoto", # 20th
 	],
 	iGreatStatesman : [
-		"Shoutouku Taishi", # 6th
+		"Shotoku Taishi", # 6th
```
dguenms: According to Wikipedia, the correct transliteration is Shoutoku Taishi. DoC consistently uses wapuro romaji for transliteration, where long o = ou.

Assets/Python/GreatPeople.py
```diff
@@ -666,15 +700,15 @@ def setup():
...
 		iMedieval,
 		"Taira no Kiyomori", # 12th
 		iRenaissance,
 		"Tokugawa Ieyasu", # 16th
 		"Arai Hakuseki", # 17th
 		iIndustrial,
-		"Sakamoto Ryouma", # 19th
-		"Oukubo Toshimichi", # 19th
+		"Sakamoto Ryoma", # 19th
```
dguenms: Ryouma (see above).

Assets/Python/GreatPeople.py
```diff
@@ -666,15 +700,15 @@ def setup():
...
+		"Okubo Toshimichi", # 19th
```
dguenms: Oukubo (see above).

Assets/Python/GreatPeople.py
```diff
@@ -2228,6 +2317,12 @@ def setup():
 		iModern,
 		"Tenzin Gyatso", # 20th
 	],
+	iGreatScientist : [
```
dguenms, line 2333: With the current code, Tibet won't use these entries, but the Chinese lists instead (there is a special rule coded into the getAlias() method above). We can change this, but I'm not sure if it's better to only have one name in your list.

dguenms (changes requested):
Overall really great changes, both all the minor bugs and misspellings you caught and the new additions and era adjustments.

Only some minor comments, we can merge after you have addressed them.

etiennefd:
Thanks for your comments! They should now be addressed. 

I have to disagree that DoC "consistently uses wapuro romaji for transliteration", since the long o's were represented by "o" in all Japanese names except the three I changed! But according to your preference (which is not the same as mine), I changed all of them to "ou". 

The only proper response to your comment on Tibetan names, of course, was do some research and make decent lists for all categories. (Some names, like Yeshe Tsogyal and Tsongkhapa, probably fit more as prophets, but I didn't want to mess up the Dalai-Lama concept.)

dguenms:
Thanks for changing the Japanese names. You're right that "consistently" is a bit of an overstatement, but it is my preferred standard (considering Civ4 cannot render the appropriate diacritic), and stuff like city names already adheres to the rule pretty strictly, so names should too.

Good call on keeping the prophets limited to dalai lamas, it's mostly that way to improve immersion when playing the Tibetan UHV. So for the moment we are agreed to keep the other Tibetan GP names, but keep them disabled, until the lists are further expanded?

etiennefd:
I think there are sufficient Tibetan names now, and in fact I took the initiative to remove the getAlias() line already. Your call.

dguenms:
Oh sorry, haven't even looked into the code again yet. That's indeed an impressive list! Merge coming.

dguenms (approved):

PR #143, opened 2017-01-20
More code cleanup and optimization

MeuhV1:
In the second commit, I simplified some functions. Some functions of them are not used at all, likely being leftovers from Rhye. Should I completely remove them?

### DIFF
--- a/Assets/Python/RFCUtils.py
+++ b/Assets/Python/RFCUtils.py
@@ -587,264 +587,180 @@ class RFCUtils:
 		return False
 
 	#Barbs, RiseAndFall
-	def squareSearch(self, tTopLeft, tBottomRight, function, argsList, tExceptions = () ): #by LOQ
+	def squareSearch(self, tTopLeft, tBottomRight, function, argsList, tExceptions = (), bOuter = False): #by LOQ
 		"""Searches all tile in the square from tTopLeft to tBottomRight and calls function for
-		every tile, passing argsList. The function called must return a tuple: (1) a result, (2) if
+		every tile, passing argsList. The function called must return a tuple: (1) a (2) if
 		a plot should be painted and (3) if the search should continue."""
-		return self.listSearch(self.getPlotList(tTopLeft, tBottomRight, tExceptions), function, argsList)
+		return self.listSearch(self.getPlotList(tTopLeft, tBottomRight, tExceptions), function, argsList, bOuter = False)
 		
-	def listSearch(self, lPlots, function, argsList):
+	def listSearch(self, lPlots, function, argsList, bOuter):
 		tPaintedList = []
-		result = None
-		for (x, y) in lPlots:
-			result, bPaintPlot, bContinueSearch = function((x, y), result, argsList)
-			if bPaintPlot:			# paint plot
-				tPaintedList.append((x, y))
-			if not bContinueSearch:		# goal reached, so stop
-				return result, tPaintedList
-		return result, tPaintedList
+		for tPlot in lPlots:
+			bPaintPlot = function(tPlot, argsList, bOuter)
+			if bPaintPlot: # paint plot
+				tPaintedList.append(tPlot)
+		return tPaintedList
 
 	#Barbs, RiseAndFall
-	def outerInvasion(self, tCoords, result, argsList):
+	def outerInvasion(self, tCoords, argsList, bOuter):
 		"""Checks validity of the plot at the current tCoords, returns plot if valid (which stops the search).
 		Plot is valid if it's hill or flatlands, it isn't marsh or jungle, it isn't occupied by a unit or city and if it isn't a civ's territory"""
-		bPaint = True
-		bContinue = True
+		return self.invasion(tCoords, argList, True)
+	
+	def invasion(self, tCoords, argList, bOuter):
 		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
 		if pPlot.isHills() or pPlot.isFlatlands():
 			if pPlot.getTerrainType() != iMarsh and pPlot.getFeatureType() != iJungle:
 				if not pPlot.isCity() and not pPlot.isUnit():
-					if pPlot.countTotalCulture() == 0:
-						# this is a good plot, so paint it and continue search
-						return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
+					if not (bOuter and pPlot.countTotalCulture() != 0):
+						return True
+		return False
 
 	#Barbs
-	def innerSeaSpawn(self, tCoords, result, argsList):
+	def innerSeaSpawn(self, tCoords, argsList): # Unused
 		"""Checks validity of the plot at the current tCoords, returns plot if valid (which stops the search).
 		Plot is valid if it's water and it isn't occupied by any unit. Unit check extended to adjacent plots"""
-		bPaint = True
-		bContinue = True
-		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
-		if pPlot.isWater():
-			if not pPlot.isUnit() and pPlot.area().getNumTiles() > 10:
-				bClean = True
-				for (x, y) in self.surroundingPlots(tCoords, 1):
-					if pPlot.getNumUnits() != 0:
-						bClean = False
-						break
-				if bClean:
-					# this is a good plot, so paint it and continue search
-					return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
+		return self.seaSpawn(tCoords, arglist, False)
 
-	#Barbs
-	def outerSeaSpawn(self, tCoords, result, argsList):
+	def outerSeaSpawn(self, tCoords, argsList): # Unused
 		"""Checks validity of the plot at the current tCoords, returns plot if valid (which stops the search).
 		Plot is valid if it's water and it isn't occupied by any unit and if it isn't a civ's territory. Unit check extended to adjacent plots"""
-		bPaint = True
-		bContinue = True
+		return self.seaSpawn(tCoords, arglist, True)
+	
+	def seaSpawn(self, tCoords, arglist, bOuter): # Used by unused functions
 		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
 		if pPlot.isWater():
 			if not pPlot.isUnit() and pPlot.area().getNumTiles() > 10:
-				if pPlot.countTotalCulture() == 0:
-					bClean = True
+				if not (bOuter and pPlot.countTotalCulture() != 0):
 					for (x, y) in self.surroundingPlots(tCoords, 1):
 						if pPlot.getNumUnits() != 0:
-							bClean = False
-							break
-					if bClean:
-						# this is a good plot, so paint it and continue search
-						return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
+							return False
+					return True
+		return False
 
-	def outerCoastSpawn(self, tCoords, result, argsList):
+	def outerCoastSpawn(self, tCoords, argsList): # Unused
 		"""Checks validity of the plot at the current tCoords, returns plot if valid (which stops the search).
 		Plot is valid if it's water and it isn't occupied by any unit and if it isn't a civ's territory. Unit check extended to adjacent plots"""
-		bPaint = True
-		bContinue = True
 		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
 		if pPlot.getTerrainType() == iCoast:
 			if not pPlot.isUnit() and pPlot.area().getNumTiles() > 10:
 				if pPlot.countTotalCulture() == 0:
-					bClean = True
 					for (x, y) in self.surroundingPlots(tCoords, 1):
 						if pPlot.getNumUnits() != 0:
-							bClean = False
-							break
-					if bClean:
-						# this is a good plot, so paint it and continue search
-						return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
+							return False
+					return True
+		return False
 
 	#Barbs
-	def outerSpawn(self, tCoords, result, argsList):
+	def outerSpawn(self, tCoords, argsList): # Unused
 		"""Checks validity of the plot at the current tCoords, returns plot if valid (which stops the search).
 		Plot is valid if it's hill or flatlands, it isn't marsh or jungle, it isn't occupied by a unit or city and if it isn't a civ's territory.
 		Unit check extended to adjacent plots"""
-		bPaint = True
-		bContinue = True
+		return self.landSpawn(tCoords, argsList, True)
+		
+	def landSpawn(self, tCoords, argsList, bOuter):
 		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
 		if pPlot.isHills() or pPlot.isFlatlands():
 			if pPlot.getTerrainType() != iMarsh and pPlot.getFeatureType() != iJungle:
 				if not pPlot.isCity() and not pPlot.isUnit():
-					bClean = True
 					for (x, y) in self.surroundingPlots(tCoords, 1):
 						if pPlot.getNumUnits() != 0:
-							bClean = False
-							break
-					if bClean:
+							return False
+					if bOuter:
 						if pPlot.countTotalCulture() == 0:
-							# this is a good plot, so paint it and continue search
-							return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
-
-	#RiseAndFall
-	def innerInvasion(self, tCoords, result, argsList):
-		"""Checks validity of the plot at the current tCoords, returns plot if valid (which stops the search).
-		Plot is valid if it's hill or flatlands, it isn't marsh or jungle, it isn't occupied by a unit or city and if it isn't a civ's territory"""
-		bPaint = True
-		bContinue = True
-		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
-		if pPlot.isHills() or pPlot.isFlatlands():
-			if pPlot.getTerrainType() != iMarsh and pPlot.getFeatureType() != iJungle:
-				if not pPlot.isCity() and not pPlot.isUnit():
-					if pPlot.getOwner() in argsList:
-						# this is a good plot, so paint it and continue search
-						return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
-
-	def internalInvasion(self, tCoords, result, argsList):
-		"""Like inner invasion, but ignores territory, to allow for more barbarians"""
-		bPaint = True
-		bContinue = True
-		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
-		if pPlot.isHills() or pPlot.isFlatlands():
-			if pPlot.getTerrainType() != iMarsh and pPlot.getFeatureType() != iJungle:
-				if not pPlot.isCity() and not pPlot.isUnit():
-					return (None, bPaint, bContinue)
-		return (None, not bPaint, bContinue)
-
-	def innerSpawn(self, tCoords, result, argsList):
-		"""Checks validity of the plot at the current tCoords, returns plot if valid (which stops the search).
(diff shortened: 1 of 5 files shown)


Assets/Python/RFCUtils.py
```diff
@@ -102,32 +79,32 @@ def isMortalUnit(self, unit):
 
 	def isDefenderUnit(self, unit):
 		iUnitType = unit.getUnitType()
-		if (iUnitType >= iSpearman and iUnitType <= iChineseChokonu):
+		if iSpearman <= iUnitType <= iChineseChokonu:
```
dguenms: Is that even still true? I would actually prefer something that depends on some of the actual attributes of the unit.

Assets/Python/RFCUtils.py
```diff
@@ -378,116 +347,76 @@ def cultureManager(self, tCityPlot, iCulturePercent, iNewOwner, iOldOwner, bBarb
...
-						elif (gc.getPlayer(iMajorCiv).getSettlerValue(x, y) >= 150):
-							iDen = 15
-							
-						iMinorCityCulture = city.getCulture(iMinor)
-						city.setCulture(iMajorCiv, iMinorCityCulture/iDen, True)
-						
-						iMinorPlotCulture = pPlot.getCulture(iMinor)
-						pPlot.setCulture(iMajorCiv, iMinorPlotCulture/iDen, True)
+	def spreadMajorCulture(self, iMajorCiv, iX, iY):
+		for (x, y) in self.surroundingPlots((iX, iY), 4):
```
dguenms: Rhye's code only has range 3.

Assets/Python/RFCUtils.py
```diff
@@ -378,116 +347,76 @@ def cultureManager(self, tCityPlot, iCulturePercent, iNewOwner, iOldOwner, bBarb
...
+									gc.getMap().plot(x, y).setCulture(iMinor, 0, True)
+									gc.getMap().plot(x, y).setCulture(iNewOwner, iMinorCulture, True)
 									
 		#plot
-		for x in range(tCityPlot[0]-1, tCityPlot[0]+2):	# from x-1 to x+1
-			for y in range(tCityPlot[1]-1, tCityPlot[1]+2):	# from y-1 to y+1
-				pPlot = gc.getMap().plot(x, y)
-				
-				iCurrentPlotCulture = pPlot.getCulture(iOldOwner)
+		for (x, y) in self.surroundingPlots(tCityPlot, 1):
```
dguenms: You can skip the second parameter, it's the default value.

Assets/Python/RFCUtils.py
```diff
@@ -545,48 +449,42 @@ def getMaster(self, iCiv):
...
-		for x in range(tCityPlot[0]-2, tCityPlot[0]+3):
-			for y in range(tCityPlot[1]-2, tCityPlot[1]+3):
-				pDestination = gc.getMap().plot(x, y)
-				if (pDestination.getOwner() == iOldOwner and (not pDestination.isWater()) and (not pDestination.isImpassable())):
-					tDestination = (x, y)
-					break
-					break
-		if (tDestination != (-1, -1)):
-			plotCity = gc.getMap().plot(tCityPlot[0], tCityPlot[1])
+		for (i, j) in self.surroundingPlots(tCityPlot, 3):
```
dguenms: Rhye's code has range 2.

Assets/Python/RFCUtils.py
```diff
@@ -545,48 +449,42 @@ def getMaster(self, iCiv):
...
 			pCity = self.getRandomCity(iOldOwner)
 			if pCity:
-				plot = gc.getMap().plot(tCityPlot[0],tCityPlot[1])
+				plot = gc.getMap().plot(x, y)
 				iNumUnits = plot.getNumUnits()
 				j = 0
 				for i in range(iNumUnits):
 					unit = plot.getUnit(j)
-					if (unit.getDomainType() == 2): #land
+					if unit.getDomainType() == 2: #land
```
dguenms: Can you use the domain type enum here? There should be something like DomainTypes.LAND

Assets/Python/RFCUtils.py
```diff
@@ -677,274 +569,180 @@ def isAVassal(self, iCiv):
 		return False
 
 	#Barbs, RiseAndFall
-	def squareSearch(self, tTopLeft, tBottomRight, function, argsList, tExceptions = () ): #by LOQ
+	def squareSearch(self, tTopLeft, tBottomRight, function, argsList, tExceptions = (), bOuter = False): #by LOQ
```
dguenms: Can we make the new bOuter parameter part of the argsList? That would save us from changing all those method signatures.

Assets/Python/RFCUtils.py
```diff
@@ -677,274 +569,180 @@ def isAVassal(self, iCiv):
...
-		pPlot = gc.getMap().plot( tCoords[0], tCoords[1] )
-		if (pPlot.getOwner() == argsList ):
-			if (pPlot.isCity()):
-				# this is a good plot, so paint it and continue search
-				return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
-
-	def ownedCityPlotsAdjacentArea( self, tCoords, result, argsList ):
+		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
```
dguenms: Unpack tCoords first?

Assets/Python/RFCUtils.py
```diff
@@ -677,274 +569,180 @@ def isAVassal(self, iCiv):
...
+				return True
+		return False
+
+	def ownedCityPlotsAdjacentArea(self, tCoords, argsList): # Unused
 		"""Checks validity of the plot at the current tCoords, returns plot if valid (which stops the search).
 		Plot is valid if it contains a city belonging to the civ"""
-		bPaint = True
-		bContinue = True
-		pPlot = gc.getMap().plot( tCoords[0], tCoords[1] )
+		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
```
dguenms: See above.

Assets/Python/RFCUtils.py
```diff
@@ -677,274 +569,180 @@ def isAVassal(self, iCiv):
...
 		#print(tCoords[0], tCoords[1], pPlot.isCity(), pPlot.getOwner() == argsList[0], pPlot.isAdjacentToArea(gc.getMap().plot(argsList[1][0],argsList[1][1]).area()))
-		if (pPlot.getOwner() == argsList[0] and pPlot.isAdjacentToArea(gc.getMap().plot(argsList[1][0],argsList[1][1]).area())):
-			if (pPlot.isCity()):
-				# this is a good plot, so paint it and continue search
-				return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
-
-	def foundedCityPlots( self, tCoords, result, argsList ):
+		if pPlot.getOwner() == argsList[0] and pPlot.isAdjacentToArea(gc.getMap().plot(argsList[1][0],argsList[1][1]).area()):
```
dguenms, line 738: See above.

Assets/Python/RFCUtils.py
```diff
@@ -677,274 +569,180 @@ def isAVassal(self, iCiv):
...
-		pPlot = gc.getMap().plot( tCoords[0], tCoords[1] )
-		if (pPlot.isCity()):
-			if (pPlot.getPlotCity().getOriginalOwner() == argsList ):
-				# this is a good plot, so paint it and continue search
-				return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
-
-	def ownedPlots( self, tCoords, result, argsList ):
+		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
```
dguenms: See above.

Assets/Python/RFCUtils.py
```diff
@@ -677,274 +569,180 @@ def isAVassal(self, iCiv):
...
-		bContinue = True
-		pPlot = gc.getMap().plot( tCoords[0], tCoords[1] )
-		if (pPlot.getOwner() == argsList ):
-			# this is a good plot, so paint it and continue search
-			return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
-
-	def goodOwnedPlots( self, tCoords, result, argsList ):
+		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
```
dguenms: See above.

Assets/Python/RFCUtils.py
```diff
@@ -677,274 +569,180 @@ def isAVassal(self, iCiv):
...
-		pPlot = gc.getMap().plot( tCoords[0], tCoords[1] )
-		if ( pPlot.isHills() or pPlot.isFlatlands() ):
-			if (pPlot.getTerrainType() != iMarsh) and (pPlot.getFeatureType() != iJungle):
-				if ( not pPlot.isCity() and not pPlot.isUnit() ):
-					    if (pPlot.getOwner() == argsList ):
-							# this is a good plot, so paint it and continue search
-							return (None, bPaint, bContinue)
-		# not a good plot, so don't paint it but continue search
-		return (None, not bPaint, bContinue)
+		pPlot = gc.getMap().plot(tCoords[0], tCoords[1])
```
dguenms: See above.

Assets/Python/RFCUtils.py
```diff
@@ -994,15 +792,13 @@ def getCoreUnitList(self, iCiv, reborn):
 
 	def getCivRectangleCities(self, iCiv, TopLeft, BottomRight):
 		cityList = []
-		for x in range(TopLeft[0], BottomRight[0]+1):
-			for y in range(TopLeft[1], BottomRight[1]+1):
-				plot = gc.getMap().plot(x,y)
-				if plot.isCity():
-					cityList.append(plot.getPlotCity())
+		for (x, y) in self.getPlotList(TopLef, BottomRight):
```
dguenms: :+1:

dguenms: Needs to be TopLeft though. Or better tTopLeft and tBottomRight.

Assets/Python/RFCUtils.py
```diff
@@ -1016,52 +812,46 @@ def removeReligionByArea(self, lPlotList, iReligion):
...
-			for j in range(y-1, y+2):
-				if (i, j) != (x, y):
-					plot = gc.getMap().plot(i, j)
-					if (not plot.isPeak()) and (not plot.isWater()) and (not plot.isCity()) and (not plot.isUnit()):
-						plotList.append((i, j))
-		iRand = gc.getGame().getSorenRandNum(len(plotList), '')
-		return plotList[iRand]
-
-	def colonialConquest(self, iCiv, x, y):
+		for (x, y) in self.surroundingPlots(tPlot, 1):
```
dguenms: Second parameter is not required.

Assets/Python/RFCUtils.py
```diff
@@ -1016,52 +812,46 @@ def removeReligionByArea(self, lPlotList, iReligion):
...
+		iTargetCiv = gc.getMap().plot(x, y).getPlotCity().getOwner()
 		lFreePlots = []
 		
-		for i in range(x-1, x+2):
-			for j in range(y-1, y+2):
-				current = gc.getMap().plot(i, j)
-				if not current.isCity() and not current.isPeak() and not current.isWater():
-					#if not current.getFeatureType() == iJungle and not current.getTerrainType() == iMarsh:
-					lFreePlots.append((i,j))
+		for (i, j) in self.surroundingPlots(tPlot, 1):
```
dguenms: See above.

Assets/Python/RFCUtils.py
```diff
@@ -1151,16 +941,11 @@ def getColonialTargets(self, iPlayer, bEmpty=False):
...
-				x, y = lPlotList[iRand]
 				bValid = True
-				for i in range(x-1, x+2):
-					for j in range(y-1, y+2):
-						if gc.getMap().plot(i, j).isCity():
-							bValid = False
-							break
-							break
-							
+				for (i, j) in self.surroundingPlots(self.getRandomEntry(lPlotList), 1):
```
dguenms: See above.

Assets/Python/RFCUtils.py
```diff
@@ -1243,14 +1028,12 @@ def getFreePlot(self, x, y):
...
 			
-		for i in range(x-1, x+2):
-			for j in range(y-1, y+2):
-				pPlot = gc.getMap().plot(i, j)
-				if not (pPlot.isCity() or pPlot.isPeak() or pPlot.isWater()):
-					lFreePlots.append((i, j))
-					
-		iRand = gc.getGame().getSorenRandNum(len(lFreePlots), 'random plot')
-		return lFreePlots[iRand]
+		for (i, j) in self.surroundingPlots((x, y), 1):
```
dguenms: See above.

Assets/Python/RFCUtils.py
```diff
@@ -1632,15 +1415,14 @@ def showPopup(self, popupID, title, message, labels):
...
 		y = city.getY()
-		for i in range(x-1, x+2):
-			for j in range(y-1, y+2):
-				plot = gc.getMap().plot(i, j)
-				if (i, j) == (x, y):
-					self.convertPlotCulture(plot, iPlayer, 25, False)
-				elif plot.getOwner() == iPreviousOwner:
-					self.convertPlotCulture(plot, iPlayer, 50, True)
-				else:
-					self.convertPlotCulture(plot, iPlayer, 25, True)
```
dguenms, line 1658: See above.

Assets/Python/RFCUtils.py
```diff
@@ -1821,9 +1603,9 @@ def moveToClosestCity(self, unit):
 		else: unit.setXY(x, y, False, True, False)
 		
 	def evacuate(self, tPlot):
-		for tLoopPlot in utils.surroundingPlots(tPlot):
+		for tLoopPlot in self.surroundingPlots(tPlot):
```
dguenms, line 1636: :+1:

Assets/Python/Victory.py
```diff
@@ -950,9 +950,8 @@ def checkTurn(iGameTurn, iPlayer):
 				lose(iRussia, 0)
 				
 		if isPossible(iRussia, 0):
-			if teamRussia.isHasTech(iRailroad):
-				if isConnectedByRailroad(iRussia, Areas.getCapital(iRussia), lSiberianCoast):
-					win(iRussia, 0)
+			if isConnectedByRailroad(iRussia, Areas.getCapital(iRussia), lSiberianCoast):
```
dguenms, line 953: Please keep this, as I recently commented in the forum that check is in to avoid unnecessary computation. Apparently it is just missing for the UHV display.

dguenms (changes requested):
I think we can scrap all unused methods that are left over from Rhye. Please do that in a separate pull request though, when other methods are extensively rewritten it's hard to tell what is being removed or only replaced.

Assets/Python/RFCUtils.py
MeuhV1: I don't know. I just simplified the code. I didn't check the functionality.

Assets/Python/RFCUtils.py
MeuhV1: Looking at the code it is 3 indeed, but the comment said 4. I think I took the value from the comment. I will change it to 3.

for x in range(iX-3, iX+4):	# from x-4 to x+4

Assets/Python/RFCUtils.py
MeuhV1: Not necessary. The bOuter is not needed, but I  forgot to remove it.

The argslist isn't really a list of input arguments it seems. It is used as a list of civs. I think it's better to change the name to civList or something and give it an empty list as default value.

Assets/Python/Victory.py
MeuhV1: I moved the railroad tech check to the isConnectedByRailroad() function. The function will return False if the player does not have the railroad tech before the computations are done. This way  these computations will also be skipped when opening the victoryscreen while the player does not have the railroad tech.

It basically fixes the victory screen without more computations.

Assets/Python/RFCUtils.py
MeuhV1: I'm not really certain what all arguments in the argslist are. I could not find an example, because this function is not used in any python file. (It will be removed soon very likely anyway.)

Assets/Python/RFCUtils.py
dguenms: Okay let's leave it out of the scope here, I've made a ticket to investigate.

Assets/Python/RFCUtils.py
dguenms: I think the whole point of using argsList is that you don't have to know what is in there. This way all anonymous functions can be implemented with the expectation of receiving only the argsList as an argument.

Assets/Python/Victory.py
dguenms: Okay, that's good.

dguenms:
Alright, from a code perspective, refactor the argsList / bOuter lines and I'm happy. Will run a couple of test starts before merging though.

Assets/Python/RFCUtils.py
```diff
@@ -1151,16 +953,11 @@ def getColonialTargets(self, iPlayer, bEmpty=False):
...
-							bValid = False
-							break
-							break
-							
+				for (i, j) in self.surroundingPlots(self.getRandomEntry(lPlotList)):
+					if gc.getMap().plot(i, j).isCity():
+						bValid = False
+						break
 				if bValid:
 					targetList.append(lPlotList[iRand])
```
dguenms, line 979: Python exception here, looks like the plot needs to be assigned to a variable after all.

Assets/Python/RFCUtils.py
```diff
@@ -1236,21 +1033,20 @@ def relocateCapital(self, iPlayer, newCapital):
 		newCapital.setHasRealBuilding(iPalace, True)
 		oldCapital.setHasRealBuilding(iPalace, False)
 		
-	def getFreePlot(self, x, y):
+	def getFreePlot(self, tPlot):
```
dguenms, line 1053: There are still invocations of the method that pass two separate coordinates, e.g. RiseAndFall line 737.

dguenms (changes requested):

Assets/Python/RFCUtils.py
```diff
@@ -53,39 +53,16 @@ def getRandomCity(self, iPlayer):
...
-			if (result):
-				return ((result.getX(), result.getY()))
+		for (x, y) in self.surroundingPlots(tPlot):
+			pPlot = gc.getMap().plot(x, y)
+			if not pPlot.isWater() and not pPlot.isPeak():
+				if not pPlot.isUnit():
+					plotList.append(pPlot)
+
+		if plotList:
+			return self.getRandomEntry(plotList)
```
dguenms: This returns a PyPlot object, not the corresponding (x, y) tuple as intended. I suggest putting the tuple into the list when it is created above right away, it never was a good idea to fill it with objects.

dguenms (changes requested):

dguenms:
Sorry, I recently refactored some stuff in Plague.py, you'll have to rebase.

MeuhV1:
Rebased and included updates

Assets/Python/RFCUtils.py
```diff
@@ -692,274 +584,192 @@ def isAVassal(self, iCiv):
 		return False
 
 	#Barbs, RiseAndFall
-	def squareSearch(self, tTopLeft, tBottomRight, function, argsList, tExceptions = () ): #by LOQ
+	def squareSearch(self, tTopLeft, tBottomRight, function, argsList, tExceptions = ()): #by LOQ
 		"""Searches all tile in the square from tTopLeft to tBottomRight and calls function for
-		every tile, passing argsList. The function called must return a tuple: (1) a result, (2) if
+		every tile, passing argsList. The function called must return a tuple: (1) a (2) if
 		a plot should be painted and (3) if the search should continue."""
 		return self.listSearch(self.getPlotList(tTopLeft, tBottomRight, tExceptions), function, argsList)
```
dguenms, line 591:
```Traceback (most recent call last):
  File "BugEventManager", line 400, in _handleDefaultEvent
  File "CvRFCEventHandler", line 578, in onBeginGameTurn
  File "RiseAndFall", line 764, in checkTurn
  File "RFCUtils", line 591, in squareSearch
TypeError: listSearch() takes exactly 5 arguments (4 given)```

dguenms (changes requested):

MeuhV1:
Fixed python exception.

dguenms (approved):

PR #148, opened 2017-02-28
CNM viewer for settler

MeuhV1:


MeuhV1:
The Mayan contact UHV now loops over less plots. (should be faster) Also added a condition that only revealed land tiles will cause the UHV to fail.

The IsConnectedByRailroad first checks if at least one of the plots in the target list is connected by railroad. If not, the heap calculations can (and now will) be skipped because they would always return False.

dguenms:
Got this in Dutch 600 AD autoplay:
```Traceback (most recent call last):
  File "BugEventManager", line 400, in _handleDefaultEvent
  File "CvRFCEventHandler", line 578, in onBeginGameTurn
  File "RiseAndFall", line 834, in checkTurn
  File "RiseAndFall", line 880, in rebirthFirstTurn
  File "RFCUtils", line 1080, in isFree
TypeError: unpack non-sequence

**Comment by MeuhV1** [comment]
Fixed

**Comment by dguenms** [comment]
Thanks, will give it another look.

File: Assets/Python/RFCUtils.py [hunk 1]
```diff
@@ -1146,26 +1138,24 @@ def moveSlaveToNewWorld(self, iPlayer, unit):
...
-				plot = gc.getMap().plot(x, y)
-				if plot.getOwner() == iPlayer:
-					if plot.getImprovementType() == gc.getInfoTypeForString("IMPROVEMENT_SLAVE_PLANTATION"):
-						plot.setImprovementType(gc.getInfoTypeForString("IMPROVEMENT_PLANTATION"))
-					if plot.isCity():
-						self.removeSlaves(plot.getPlotCity())
+		for (x, y) in self.getWorldPlotsList():
+			plot = gc.getMap().plot(x, y)
+			if plot.getOwner() == iPlayer:
+				if plot.getImprovementType() == iPlantation:
```
dguenms: You mixed up the plantations here, slave plantations are supposed to be converted to regular plantations.

Assets/Python/RFCUtils.py
```diff
@@ -1645,15 +1631,15 @@ def moveToClosestCity(self, unit):
...
+			for unit in self.getUnitList(tLoopPlot):
+				lPossibleTiles = self.surroundingPlots(tLoopPlot, 2, lambda (x, y): self.isFree(unit.getOwner(), (x, y), bNoEnemyUnit=True, bCanEnter=True))
+				tTargetPlot = self.getRandomEntry(lPossibleTiles)
 				if tTargetPlot:
 					x, y = tLoopPlot
 					unit.setXY(x, y, False, True, False)
 			
 	def getWonderList():
-		return [i for i in range(iNumBuildings) if isWorldWonderClass(gc.getBuildingInfo(i).getBuildingClassType())]
+		return [i for i in range(iBeginWonders, iNumBuildings)]
```
dguenms: I prefer the implementation using `isWorldWonderClass` actually, it does not depend on the `iBeginWonders` constant being set correctly.

dguenms (changes requested):
Only two minor changes.

MeuhV1:
Done.

dguenms (approved):

PR #153, opened 2017-04-21
Actual Specialist info (BUG) takes happiness changes into account

MeuhV1:


CvGameCoreDLL/CvGameTextMgr.cpp
```diff
@@ -6492,6 +6492,19 @@ void CvGameTextMgr::parseSpecialistHelpActual(CvWStringBuffer &szHelpString, Spe
 			}
 			bStarted = setResumableCommerceTimes100ChangeHelp(szHelpString, szStart, L": ", L"", aiCommerces, true, bStarted);
 
+			// Happiness
+			if (iHappinessChange != 0)
+			{
+				if (iHappinessChange > 0)
```
dguenms: You can remove that `if` statement and collapse the two `setResumableValueChangeHelp` calls (see other examples that display happiness), something like `setResumableValueChangeHelp(szHelpString, szStart, L": ", L"", std::abs(iHappinessChange), iHappiness > 0 ? gDLL->getSymbolID(HAPPY_CHAR) : gDLL->getSymbolID(UNHAPPY_CHAR), false, true, bStarted)`

dguenms (changes requested):
Great idea, I'll also look at it within the game before approving.

MeuhV1:
New code, same change

CvGameCoreDLL/CvGameTextMgr.cpp
```diff
@@ -6492,6 +6492,9 @@ void CvGameTextMgr::parseSpecialistHelpActual(CvWStringBuffer &szHelpString, Spe
 			}
 			bStarted = setResumableCommerceTimes100ChangeHelp(szHelpString, szStart, L": ", L"", aiCommerces, true, bStarted);
 
+			// Happiness
+			bStarted = setResumableValueChangeHelp(szHelpString, szStart, L": ", L"", std::abs(iHappinessChange), iHappinessChange > 0 ? gDLL->getSymbolID(HAPPY_CHAR) : gDLL->getSymbolID(UNHAPPY_CHAR), false, true, bStarted);
```
dguenms, line 6496: You still need the `iHappiness != 0` check though.

dguenms (changes requested):

CvGameCoreDLL/CvGameTextMgr.cpp
MeuhV1: There already is such a check in the setResumableValueChangeHelp function itself, so I could leave it out. Tested it and it works.

dguenms: Alright!

PR #154, opened 2017-04-28
Various small fixes

MeuhV1:


Assets/XML/Units/CIV4UnitInfos.xml
```diff
@@ -11106,12 +11106,6 @@
 					<PassableTech>TECH_OPTICS</PassableTech>
 				</TerrainPassableTech>
 			</TerrainPassableTechs>
-			<FeaturePassableTechs>
```
dguenms, line 11109: True, but then you also need to remove the `FeatureImpassables` entry for capes.

dguenms (changes requested):

Assets/XML/Units/CIV4UnitInfos.xml
MeuhV1: True. Updated.

dguenms (approved):

PR #158, opened 2017-07-15
Various updates

MeuhV1:
If year is after company fall date and that company is removed from all cities, it skips the cityValue calculations and the spreading/disappearing of companies
Improved canRespawn() function in RFCUtils by adding more return statements
Added pedia jumps to routes page in Improvements and Resources pedia
Removed sensibility check in WB DoC maps editor (was not used anyway)
Some other small fixes/updates

### DIFF
--- a/Assets/Python/Pedia/CvPediaImprovement.py
+++ b/Assets/Python/Pedia/CvPediaImprovement.py
@@ -193,7 +193,7 @@ class CvPediaImprovement:
 				if iYieldChange > 0:
 					szSign = u"+"
 				if iYieldChange != 0:
-					szText += u"%c%s%d%c with <link=CONCEPT_MOVEMENT>%s</link>\n" % (szBullet, szSign, iYieldChange, gc.getYieldInfo(iYieldType).getChar(), gc.getRouteInfo(iRoute).getDescription())
+					szText += u"%c%s%d%c with <link=literal>%s</link>\n" % (szBullet, szSign, iYieldChange, gc.getYieldInfo(iYieldType).getChar(), gc.getRouteInfo(iRoute).getDescription())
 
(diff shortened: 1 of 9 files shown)


Assets/Python/Pedia/CvPediaImprovement.py
```diff
@@ -193,7 +193,7 @@ def placeDetails(self):
 				if iYieldChange > 0:
 					szSign = u"+"
 				if iYieldChange != 0:
-					szText += u"%c%s%d%c with <link=CONCEPT_MOVEMENT>%s</link>\n" % (szBullet, szSign, iYieldChange, gc.getYieldInfo(iYieldType).getChar(), gc.getRouteInfo(iRoute).getDescription())
```
dguenms, line 196: 👍

Assets/Python/RFCUtils.py
```diff
@@ -1573,25 +1572,26 @@ def canRespawn(self, iPlayer):
 			for tInterval in tResurrectionIntervals[iPlayer]:
 				iStart, iEnd = tInterval
 				if getTurnForYear(iStart) <= iGameTurn <= getTurnForYear(iEnd):
-					bPossible = True
 					break
+			else:
```
dguenms, line 1576: ... is that an else block after a for loop?

MeuhV1: Yes. Python has a else clause for for loops.
http://book.pythontips.com/en/latest/for_-_else.html

The else activates when the for loops completes normally. If the loop is terminated with a break statement, the else clause does not activate.

dguenms: That's really cool, I learned something today. And it's compatible with Python 2.4.

PR #170, opened 2017-10-12
Improved Great People III

etiennefd:
Tons of new Great People suggestions, including: 
- Polynesia (new)* 
- Moors (new, separate from Arabs)
- Arab names to compensate for the loss of Andalucian/Maghrebi names
- Most Roman names are now full names (e.g. Marcus Tullius Cicero instead of Cicero)
- Special focus on England, Russia, Turkey, Poland
- Miscellaneous other additions and revisions

Wiki links for every addition [here](https://www.evernote.com/shard/s419/sh/dd5568d3-c4e5-488d-bc72-db33db923a2e/051e133efc3bed0cdc9ddaabe08eb4e9)

*One particularity of Polynesia is that I put legendary navigators as Great Artists, since GAs are instrumental in discovering new islands with culture, and I couldn't find any pre-contact GAs otherwise.


dguenms (changes requested):
Looks great, especially that we have Polynesian and Moorish names now. The only change I would request is about the English GPs, I know people like Wellington are hard to handle because they are more well known by their title, but having it like this is just too convoluted. Using their title like a last name is technically wrong but better fits with all other names in the game.

etiennefd:
Sure. I wasn't sure about how to handle them, but in the end I think I let my bias in favor of long aristocratic names show haha. How about "Arthur Wellesley Wellington" and "John Churchill Marlborough"?

dguenms (approved):

PR #171, opened 2017-10-18
Tech Advisor update & a bugfix

MeuhV1:
New button in tech advisor that shows all techs that could be bulbed by a GP

Fixed renaming cities when moving them with WB


Assets/Python/Screens/CvTechChooser.py
```diff
@@ -870,7 +869,8 @@ def placeGreatPeople(self):
 						iMaxFlavor = gc.getTechInfo(iLoopTech).getFlavorValue(iFlavor)
 						iTech = iLoopTech
 
-			screen.addDDSGFCAt("GreatPeopleUnit" + str(iUnit),"TechBottomPanel", gc.getUnitInfo(iUnit).getButton(), iGPX, 16, self.GP_ICON_SIZE, self.GP_ICON_SIZE, WidgetTypes.WIDGET_PEDIA_JUMP_TO_UNIT, iUnit, -1, False)
```
dguenms: Don't we want to keep the ability to jump to the GP entry when clicking the button?

dguenms (changes requested):

Assets/Python/Screens/CvTechChooser.py
MeuhV1: That was also one of my concerns. I replaced that because the old tech screen didn't have a pedia jump at all. I'll try to come up with a solution. (I actually also prefer to have the pedia jump available)

dguenms: Okay good. It looks like it should work as it was though, what is the issue there?

MeuhV1:
Changed the tech advisor.
- Restored the jump to GP pedia for the GP button
- Added a new button for each GP that shows all techs that can be bulbed

dguenms:
Looks great!

dguenms (approved):

PR #173, opened 2017-11-03
Canada leaderhead changes

etiennefd:
- new MacDonald leaderhead (credit to Novator)
- Canada's national anthem as MacDonald's and Trudeau's diplomacy music (credit to the United States Navy Band; recording is in the public domain)
- includes all my GP names commits because I'm a noob at github branching


dguenms (changes requested):
The extra commits should be fine, I will squash all the commits into one when merging so they'll disappear. But in the future you can avoid that by branching off first: `git checkout -b <branch name>` and then creating your pull request from this branch instead of develop. 

Also, you included a bunch of `.DS_Store` files in your commits, can you remove them?

etiennefd:
I removed the .DS_Store files. 
(Also, I did create a new branch, but from my develop branch which already had the GP changes. Wasn't sure how to go around this.)
(Also, yes, I'm steb)

dguenms:
Thanks for removing those files.

You can also branch off of earlier commits by first going to the earlier commit with `git checkout <commit id>`. You can also do something like `git checkout HEAD~n` to go back `n` commits.

I recommend always using branches and using develop only to pull the latest stuff from my main repository.

etiennefd:
Thanks for the tip.

dguenms:
Viewed everything in game and it looks good. Merged.

PR #176, opened 2017-11-15
Tech cost adjustments

temujin2k:
Hello, Leo!

I tried my hand on fixing the tech costs as it became too fast since the recent commits concerning the tech tree. I've already tried a few 600 AD games with it, and global tech speed seems to have gotten back to normal, at least in the Medieval and Renaissance Eras. Haven't gotten beyond column 12 yet, though. I also haven't tried 3000 BC games yet so I'm not sure if I accidentally slowed down early game tech speed or not.

I'll encourage the others in the CFC Forums to try this out too so that we can have more input on this. Thanks!


dguenms (approved):
Looks like a reasonable increase. I guess we can best collect feedback if I merge this and adjust based on future feedback if needed.

PR #179, opened 2017-12-06
Persia no longer spawns at war against India

Alexius08:
https://forums.civfanatics.com/threads/suggestions-and-requests.533868/page-209

Not much objections were raised for this one.


dguenms:
Yeah, I also agreed. Thanks, this saves me the effort.

PR #181, opened 2017-12-10
Map data fixes

Alexius08:
- Multi-era name updates now enabled. They will now work when building and acquiring cities. This affects China the most, since many of their cities get renamed every era. I'd like to have someone knowledgeable on previous Chinese city names check my work.
- Many suggested city names in Google sheets are now implemented. Some previous names are retained to maintain backwards compatibility.
- I've also added several city names for many other civs.
- Religion-based name updates for Jerusalem added.
- City names for 600 AD and 1700 AD scenarios have been updated.
- The rename dictionaries have been mostly-sorted in alphabetical order. Fully-sorting them in alphabetical order requires entering accented characters as themselves (which I've been unable to do, since they're stored in a mix of HTML entities and Windows 1252 in the WBSave files).
- I also edited SettlerMaps.py to encourage Polish settlement of Courland colonies (but not enough to make it stable territory for them) and make Philippines a stable area for America.


dguenms:
Hey, this is a great update, but way too large to properly review all at once.

First of all, I don't know if I'm ready yet to approve changed city name maps. Ideally I want to review the suggestions in Google docs and either approve or reject them there, so the finalised map can be imported into the mod.

But even then, I would prefer to review them in smaller sets of maybe one pull request per language.

Can you change the scope of your branch so that it only includes the other code changes, and not the city name maps?

Alexius08:
By "other code changes", you mean changes I made to other files and edits outside of the city name arrays and rename dicts?

dguenms:
Exactly. Ideally, you could do a separate pull request for each of the separate features you mentioned in the bullet points above. Then later, add additional city names separately by language.

Alexius08:
I finally figured out what to do.

I'll make a new branch for the following changes:
- multi-era city renames
- Communist city renames
- tertiary languages
- rename dictionary rearrangement

Each civ-specific city name map would then branch out from the main CityNameManager branch.

PR #182, opened 2017-12-22
First of a series of commits for CityNameManager and related files

Alexius08:
Here are the contents of this pull request:
- multi-era city renames
- Communist city renames
- tertiary languages
- ~rename dictionary rearrangement~ moved to #184

Hopefully, it's small enough for review.


dguenms:
Can you split off the rearrangement as well? It would make it easier to have a pull request where only arrangement changes but the effect remains the same.

Alexius08:
I've split off the sorted rename dicts to a pull request you've already approved. Are there other issues I need to address?

dguenms:
Sorry, the PR itself is fine but I need to wrap my mind around your code changes. I think some of what you did should be implemented as a function instead. I'll give more specific feedback later.

PR #183, opened 2017-12-24
Revise guide on stability in Civilopedia

Alexius08:
Based on my reading of Stability.py, I have updated the in-game documentation of the stability system.


Assets/XML/Text/Civilopedia.xml
```diff
@@ -3521,30 +3521,26 @@ Overextension is the main factor of expansion stability. It depends on two score
...
 [NEWLINE][BOLD]ECONOMY[\BOLD]
 [NEWLINE][BOLD]Economic Growth[\BOLD]
-The economic growth score is updated every third turn, in which the total amount of commerce produced is compared to the amount from three turns earlier. Has it grown by more than 5%, the score increases, has it shrunk by more than 5%, it decreases. Otherwise the score slowly moves towards 0 again. At every stability check the score is directly applied to the economy stability. Some of your civics influence your economic growth stability:
-[ICON_BULLET]you receive extra points during economic growth when running Free Market
-[ICON_BULLET]you receive no penalties during economic stagnation when running Environmentalism
-[ICON_BULLET]you receive less penalties during economic decline when running Public Welfare
+The economic growth score is updated every third turn, in which the total amount of commerce produced is compared to the amount from three turns earlier. Has it grown by more than 5%, the score increases, has it shrunk by more than 5%, it decreases. Otherwise the score slowly moves towards 0 again. At every stability check the score is directly applied to the economy stability.
```
dguenms, line 3541: That's not accurate. Three turn intervals of more than 5% growth are treated as positive growth, 0-5% growth are neutral growth, less than 0% is negative growth. When updating stability, the last ten records of economic growth are considered. The resulting stability is the difference between positive growth records and negative growth records. However, the result is subtracted by half the number of neutral growth records. 

For example:
* 6 positive growth and 4 negative growth = +2 stability
* 6 positive growth and 4 neutral growth = +4 stability
* 2 positive growth and 8 neutral growth = +0 stability (neutral can never lead to negative stability)
* 4 positive growth, 4 neutral growth, 2 negative growth = 4 - 4/2 - 2 = +0 stability

I don't think it's easy to put into words and more confusing than illuminating if described in all detail. Instead I would focus on the important aspects:
* growth is recorded every three turns
* growth can either be positive (>5%), neutral (0-5%) and negative (<0%)
* the last ten growth records are considered when economic stability is calculated
* more negative growth records lead to negative stability, more positive growth records lead to positive stability
* records of neutral growth reduce the score towards zero

Assets/XML/Text/Civilopedia.xml
```diff
@@ -3557,8 +3553,7 @@ For religions, two or three percentages are calculated:
 [ICON_BULLET]the percentage of cities that have at least one non-state religion (heathen ratio)
 [ICON_BULLET]if you are running Theocracy, the percentage of cities that have ONLY your state religion
 [NEWLINE]Some non-state religions are ignored for the purposes of the second and third ratios ("tolerance").
-[ICON_BULLET]civs with Secularism tolerate all non-state religions
-[ICON_BULLET]Mughals tolerate all non-state religions because of their UP
```
dguenms, line 3561: The Mughal UP is still in effect.

dguenms (changes requested):

Assets/XML/Text/Civilopedia.xml
Alexius08: You mean the old Mughal UP of religious tolerance? There's a new Mughal UP (earning culture points from building production) since the old one became accessible to other civs as the Tolerance civic.

dguenms: Oh right ... I should be better at remembering my changes.

Assets/XML/Text/Civilopedia.xml
Alexius08: I only removed the three bullet points and the sentence that came before that. The rest were not changed. Should I revise this section according to what you wrote here?

dguenms: Sorry, it looked like you introduced that sentence. It's up to you, if you don't want to address these inaccuracies we can leave it unchanged.

dguenms (approved):

PR #190, opened 2018-01-12
Code refactoring and bugfix in Stability.py

MeuhV1:


Assets/Python/Stability.py
```diff
@@ -690,37 +690,35 @@ def calculateStability(iPlayer):
...
+					iTempCulture *= 2
+				iTotalCulture += iTempCulture
+				
+			if iTotalCulture != 0:
+				iCulturePercent = 100 * iOwnCulture / iTotalCulture
+			else:
+				iCulturePercent = 100
+					
+			bExpansionExceptions = ((bHistorical and iPlayer == iMongolia) or bTotalitarianism)
+
```
MeuhV1, line 721: This code is only required for non-core cities. I moved it so the calculations are skipped for core cities. (I did a check and the variables aren't used anywhere else in the code.

Assets/Python/Stability.py
```diff
@@ -760,23 +758,23 @@ def calculateStability(iPlayer):
...
+						bNonStateReligion = True
+						break
+					
+			if city.isHasReligion(iStateReligion):
+				iStateReligionPopulation += iPopulation
+				if not bNonStateReligion: iOnlyStateReligionPopulation += iPopulation
+					
+			if bNonStateReligion: 
+				if iStateReligion >= 0 and city.isHasReligion(iStateReligion): iDifferentReligionPopulation += iPopulation / 2
+				else: iDifferentReligionPopulation += iPopulation
```
MeuhV1, line 778: It now skippes code if the city does not have a religion.

Minor note: I a civ has a pagan state religion, the variable iStateReligionPopulation now doesn't increase anymore. This variable is only relevant if the civ does not have a pagan state religion, so this does minor change does not matter.

Assets/Python/Stability.py
```diff
@@ -1538,13 +1533,12 @@ def getResurrectionCities(iPlayer, bFromCollapse = False):
...
-			# cities in core are not affected by this
-			for city in lPotentialCities:
-				if not city.plot().isCore(city.getOwner()):
-					lFlippingCities.append(city)
+	# if only up to two cities wouldn't flip, they flip as well (but at least one city has to flip already, else the respawn fails)
+	if len(lFlippingCities) + 2 >= len(lPotentialCities) and len(lFlippingCities) > 0 and len(lFlippingCities) * 2 >= len(lPotentialCities) and not bFromCollapse:
+		# cities in core are not affected by this
+		for city in lPotentialCities:
+			if not city.plot().isCore(city.getOwner()) and city not in lFlippingCities:
+				lFlippingCities.append(city)
```
MeuhV1, line 1542: This is the bugfix.

First, this part was inside the for-loop starting in line 493.

Secondly, if this part is executed, it was possible that some cities that were included in the first list didn't get included in the new lFlippingCities list. And the other way around. Now, the old list is maintained and is extended with new cities.

Assets/Python/Stability.py
dguenms: Good idea. However I would still prefer to have the variable initialisations at the beginning of the loop. Otherwise there will be an exception if we ever decide to use them outside of the if clause. I'm talking about `iModifier = 0` and `iTotalCulture = 100` (the latter would also remove the need for that one else clause).

Assets/Python/Stability.py
dguenms: I agree, iStateReligionPopulation does not matter without a real state religion.

dguenms (changes requested):
Only that minor refactoring, otherwise ready to merge.

MeuhV1:
Recommited the changes with the old mentioned variable initialisation positions

dguenms (approved):

PR #191, opened 2018-01-13
Fixed anarchy bug

MeuhV1:


dguenms (commented):
This still affects the hover tool tip, right? Could you previously hover over the text even during anarchy before?

MeuhV1:
The error prevented the text to appear at all, because an argument of the setText() function was missing. So I don't know if you could hover over it.

dguenms (approved):

PR #192, opened 2018-01-17
Code refactoring

MeuhV1:
Reversed looping over units. See the changes in plague.py for example

Rewrote the flipUnitsInCityBefore(), flipUnitsInCityAfter() and flipUnitsInArea() functions. It doesn't temporarily places units on (0, 67) anymore.

Deleted tTempFlippingCity from the StoredData. Temporarily saving the value is not needed.

### DIFF
--- a/Assets/Python/RFCUtils.py
+++ b/Assets/Python/RFCUtils.py
@@ -113,6 +113,7 @@ class RFCUtils:
 						teamMinor.makePeace(iActiveCiv)
 						if bOpenBorders:
 							teamMinor.signOpenBorders(iActiveCiv)
+	
 	#AIWars
 	def restorePeaceHuman(self, iMinorCiv, bOpenBorders): 
 		teamMinor = gc.getTeam(gc.getPlayer(iMinorCiv).getTeam())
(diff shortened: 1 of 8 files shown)


Assets/Python/RFCUtils.py
```diff
@@ -243,61 +240,43 @@ def killUnitsInArea(self, iPlayer, lPlots):
...
-
-
+							unit.kill(False, iBarbarian)
+							
+							# Leoreth: can't flip naval units anymore
+							if unit.getDomainType() == DomainTypes.DOMAIN_SEA:
+								continue
+								
+							# Leoreth: ignore workers as well
+							if unit.getUnitType() in [iWorker, iPunjabiWorker, iMadeireiro]:
```
dguenms: Maybe an opportunity to use utils.getBaseUnit(), also should include the Labourer unit.

dguenms (changes requested):

MeuhV1:
Update the commit with your suggestion.

Included also the following in the meantime:
- An isPossible() check in the 3rd Canadian UHV
- A bugfix for the National Theatre. (Happpiness got increased when the building was removed, instead of decreased) This has a DLL update.

I rebased all files to the latest version just before I made the changes, so that shouldn't cause any troubles. (Unless you made a change in the meantime XD)

MeuhV1:
Updated again to match the latest commit. (https://github.com/dguenms/Dawn-of-Civilization/commit/b93a5bf63de3af12bbc48716e46047f70acc5666)

Also did a minor update in CvScreenInterface.py

dguenms (approved):

PR #194, opened 2018-02-13
Minor changes

MeuhV1:
Updated strategy texts
Fixed python exception in WB stability maps editor
Very minor code refactoring
Incorporated GP naming in WB unit placing

### DIFF
--- a/Assets/Python/GreatPeople.py
+++ b/Assets/Python/GreatPeople.py
@@ -65,7 +65,7 @@ def getName(unit):
 	
 	return utils.getRandomEntry(lAvailableNames)
 	
-def onGreatPersonBorn(unit, iPlayer, city):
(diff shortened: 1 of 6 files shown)


Assets/Python/GreatPeople.py
```diff
@@ -65,7 +65,7 @@ def getName(unit):
 	
 	return utils.getRandomEntry(lAvailableNames)
 	
-def onGreatPersonBorn(unit, iPlayer, city):
+def onGreatPersonBorn(unit, iPlayer, city, bWB = False):
```
dguenms: Can you rename the new argument to something that is meaningful to the method itself? I would prefer something like `bAnnounceBirth = True`.

dguenms (changes requested):

MeuhV1:
Updated PR with your comment.

dguenms (approved):

PR #197, opened 2018-03-09
Added an option to disable Human raze penalty

MeuhV1:


dguenms:
Thank you for the pull request, but I don't think I want to offer this as an option. It is a general principle that stability rules are the same across all difficulties and there should be no option to apply them selectively.

MeuhV1:
I had a feeling you might say something like this. I even considered asking if you would include this idea before I coded it. It was only 5 minutes coding anyway.

The commit includes a minor typo fix in a strategy text. You might want to include it yourself.

dguenms:
Yeah, still sorry for your wasted efforts. I will fix the text.

PR #199, opened 2018-03-17
Improved Great People 5

etiennefd:
Highlights: 

- Special focus on China, India, Persia, Ethiopia, England

- Merged Persia and Iran 

- Great Spies
The following civs have received a fairly extensive Spy list (at least 6 names): India, Japan, Vikings, Spain, France, England, Germany, Russia, Poland, Italy, USA. 
The following civs have received 1-5 Spy names: China, Korea, Arabia, Portugal, Mughals, Turkey, Netherlands, Argentina, Mexico, Canada
The other civs have no Great Spies, either because I couldn't find any or because they're too unlikely to produce Great Spies in game (e.g. they're too early).

### DIFF
--- a/Assets/Python/GreatPeople.py
+++ b/Assets/Python/GreatPeople.py
@@ -316,7 +316,7 @@ iCivIndia : {
 		iMedieval,
 		"Aryabhata", # 6th
 		"Dignaga", # 6th
-		"Dharmakirti" # 6th or 7th
+		"Dharmakirti", # 6th or 7th
 		"Brahmagupta", # 7th
 		"Bhaskara", # 12th
 		"Madhava", # 14th
@@ -350,7 +350,7 @@ iCivIndia : {
 		"Jagadish Chandra Bose", # 19th
 		"Mokshagundam Visvesvaraya", # 19th
 		iGlobal,
-		"fEulie Chowdhury" # 20th
+		"fEulie Chowdhury", # 20th
 		"Satish Dhawan", # 20th
 		"Charles Correa", # 20th
 	],
@@ -389,7 +389,7 @@ iCivIndia : {
 		"fRani Lakshmibai", # 19th
 		iGlobal, 
 		"Kodandera M. Cariappa", # 20th
-		"Sam Manekshaw" # 20th
+		"Sam Manekshaw", # 20th
 	],
 	iGreatSpy : [
 		"Bahirji Naik", # 17th
@@ -941,7 +941,7 @@ iCivJapan : {
 		"Kenkichi Iwasawa", # 20th
 	],
 	iGreatMerchant : [
-		"Ruson Sukezaemon" # 16th
+		"Ruson Sukezaemon", # 16th
 		"Shousuke Tanaka", # 17th
 		"Takadaya Kahei", # 18th
 		iIndustrial,
--- a/Assets/Python/GreatPeople.py
+++ b/Assets/Python/GreatPeople.py
@@ -911,7 +911,8 @@ iCivJapan : {
 		"Takuan Souhou", # 17th
 		iIndustrial,
 		"Hirata Atsutane", # 19th
-		"Uchimura Kanzou", # 19th
+		"fMiki Nakayama", # 19th,
+		"Kanzou Uchimura", # 19th
 	],
 	iGreatArtist : [
 		"fMurasaki Shikibu", # 10th
@@ -941,7 +942,9 @@ iCivJapan : {
 		"Kenkichi Iwasawa", # 20th
 	],
 	iGreatMerchant : [
+		"Ruson Sukezaemon" # 16th
 		"Shousuke Tanaka", # 17th
+		"Takadaya Kahei", # 18th
 		iIndustrial,
 		"Torakusu Yamaha", # 19th
 		"Outano Kouzui", # 19th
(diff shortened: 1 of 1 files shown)


Assets/Python/GreatPeople.py
```diff
@@ -275,40 +336,70 @@ def setup():
 		"Jamshetji Tata", # 19th
 		"Ardeshir Godrej", # 19th
 		iGlobal,
+		"V. O. Chidambaram Pillai", # 20th
```
dguenms: Can you find a form of the name that isn't too long without initials?

Assets/Python/GreatPeople.py
```diff
@@ -275,40 +336,70 @@ def setup():
...
 		"Chandragupta Maurya", # 4th BC
 		"Samudragupta", # 4th BC
 		iMedieval,
-		"Rajaraja Chola", # 12th
+		"Dhruva Dharavarsha", # 8th
+		"Mihira Bhoja", # 9th
+		"Rajaraja Chola", # 10th
 		iRenaissance,
 		"fRani Durgavati", # 16th
-		"Shivaji Bhosle", # 17th
```
dguenms: Why did you remove him? The LH is Shahuji.

Assets/Python/GreatPeople.py
```diff
@@ -486,7 +577,6 @@ def setup():
 		"Fotis Kafatos", # 20th
 	],
 	iGreatMerchant : [
-		"Kroisos", # 6th BC
```
dguenms: Why remove him?

Assets/Python/GreatPeople.py
```diff
@@ -606,59 +696,146 @@ def setup():
 },
 iCivPersia : {
 	iGreatProphet : [
-		"Zoroaster", # 18-10th BC
+		"Mahabad", # legendary
+		"Zarathustra", # 18-10th BC
```
dguenms, line 701: 👍

Assets/Python/GreatPeople.py
```diff
@@ -606,59 +696,146 @@ def setup():
...
 	],
 	iGreatScientist : [
+		"Ktesias", # 5th BC
 		"Ardashir", # 4th
+		"Borzuya", # 6th
+		"Paulos-e irani", # 6th
 		iMedieval,
-		"Al-Khwarizmi", # 9th
-		"Al-Razi", # 9th
+		"Muhammad ibn Musa al-Khwarizmi", # 9th
```
dguenms: If the "last name" is well known I think it's best to leave it at that.

Assets/Python/GreatPeople.py
```diff
@@ -606,59 +696,146 @@ def setup():
...
+		iGlobal,
+		"Manny Mashouf", # 20th
+		"Nasser David Khalili", # 20th
+		"Omid Kordestani", # 20th
+		"Amir Ansari", # 20th
+		"Pierre Omidyar", # 20th
 	],
 	iGreatEngineer : [
-		"Artaxerxes", # 5th BC
+		u"Artaxsaça", # 4th BC
```
dguenms: I would prefer the more phonetic transliteration "Artakhshathra" (see [here](http://www.jewishencyclopedia.com/articles/1827-artaxerxes-i))

Assets/Python/GreatPeople.py
```diff
@@ -606,59 +696,146 @@ def setup():
...
+		iGlobal,
+		"Reza Shah Pahlavi", # 20th
+		"Mohammad Mossadegh", # 20th
 	],
 	iGreatGeneral : [
-		"Achaemenes", # 7th BC
-		"Xerxes", # 5th BC
-		"Mehrdad", # 2nd BC
+		"Haxamanis", # 7th BC
+		"Xsaya-rsa", # 5th BC
```
dguenms: Likewise, "Khashayarsha".

Assets/Python/GreatPeople.py
```diff
@@ -1641,16 +1940,20 @@ def setup():
...
 		iGlobal,
+		"Thomas Edward Lawrence", # 20th
 		"fEmmeline Pankhurst", # 20th
 		"Clement Atlee", # 20th
+		"fDiana Spencer", # 20th
 	],
 	iGreatGeneral : [
 		"William the Conqueror", # 11th
 		"Richard the Lionheart", # 12th
+		"William Wallace", # 13th
```
dguenms: I don't think Scottish figures before unification should be included (especially those who fought against England)

Assets/Python/GreatPeople.py
```diff
@@ -2487,6 +2868,14 @@ def setup():
 		iIndustrial,
 		"Henri Winkelman", # 20th
 	],
+	iGreatSpy : [
+		"fJohanna Brandt", # 19th
+		"Christiaan Snouck Hurgronje", # 19th
+		iGlobal,
+		"fMata Hari", # 20th
+		"Dirk Klop", # 20th
+		"François van 't Sant", # 20th
```
dguenms: That should be a unicode string.

dguenms (changes requested):
Great additional names as always, only some small change requests.

Assets/Python/GreatPeople.py
etiennefd: Valliappan Olaganathan Chidambaram Pillai is a little long, isn't it? Apparently he's also known as Kappalottiya Tamizhan, "The Tamil Helmsman", so I'll put that.

Assets/Python/GreatPeople.py
etiennefd: Interesting. The XML tags say Shivaji, which is what I had checked, plus the fact that Shivaji seemed like a more important figure. The civilopedia also has a piece on Shivaji, not on Shahuji. The XML text however is Shahuji, and so is the constant in Consts.py. I'll put Shivaji back in the GP list, but what is the intention here?

Assets/Python/GreatPeople.py
etiennefd: I use Wikipedia titles as a heuristic to choose the best name to use, unless that name is not in the native language of the civ, but in this case I'm reverting it.

Assets/Python/GreatPeople.py
etiennefd: Good catch.

dguenms (approved):

PR #200, opened 2018-03-21
Fixed tech bar

MeuhV1:
A very minor side effect is that you can change the current research by clicking on the bar or text, instead of only the text. But I don't think that is a problem.


dguenms:
I remember trying the same and running into the same drawback. I will look at it inside the game before merging.

PR #201, opened 2018-03-23
Khosrow

etiennefd:
This branch adds Khosrow I as a leader for Persia, representing the Sasanian/Sassanid dynasty. Art is from [here](https://forums.civfanatics.com/resources/lh-khosrau-i.13963/) and is, I think, of acceptable quality, although the sleeves look a bit weird. 

Khosrow has Frederick the Great's personality, Culture and Admin flavors (the Sasanian society was a golden age of art and had an efficient bureaucracy), Zoroastrianism as his favorite religion, and Clergy as his favorite civic (I read that the Zoroastrian clergy played a large role in Sasanian society.) 

Persia gets this leaderhead when they reach the last column of the classical era. In practice, this happens some time in the 1st or 2nd century AD, which I think is appropriate. I also changed the criterion for the Sassanid Empire dynamic name to match this (instead of starting in the medieval era), though in practice the two changes don't occur at the same time. (It's never been clear to me when dynamic names change after the criteria are met.)

Also includes a civilopedia text that I wrote based on Wikipedia.


Assets/Python/DynamicCivs.py
```diff
@@ -1283,7 +1283,7 @@ def specificAdjective(iPlayer):
 					
 				return "TXT_KEY_CIV_PERSIA_ACHAEMENID"
 			
-			if iEra == iMedieval:
+			if getColumn(iPlayer) >= 6:
```
dguenms, line 1286: :+1:

dguenms (approved):

PR #209, opened 2018-05-01
Improvements in Victory.py

MeuhV1:
Removed double function
Discover first x era technology goals fail if target is not met and all techs from that era are discovered
Discover first x era technology goals only checks when the discovered era is the era for the UHV
Tamil trade UHV stores decimal values
Fixed Enter first era goals

### DIFF
--- a/Assets/Python/Victory.py
+++ b/Assets/Python/Victory.py
@@ -538,15 +538,15 @@ def checkTurn(iGameTurn, iPlayer):
 			iTradeCommerce = 0
 			for city in utils.getCityList(iTamils):
 				iTradeCommerce += city.getTradeYield(2)
-			iTradeGold += iTradeCommerce * pTamils.getCommercePercent(0) / 100
+			iTradeGold += iTradeCommerce * pTamils.getCommercePercent(0)
 			
 			# gold from per turn gold trade
 			for iPlayer in range(iNumPlayers):
-				iTradeGold += pTamils.getGoldPerTurnByPlayer(iPlayer)
+				iTradeGold += pTamils.getGoldPerTurnByPlayer(iPlayer) * 100
 			
 			data.iTamilTradeGold += iTradeGold
 			
-			if data.iTamilTradeGold >= utils.getTurns(4000):
+			if data.iTamilTradeGold / 100 >= utils.getTurns(4000):
 				win(iTamils, 2)
 				
 		if iGameTurn == getTurnForYear(1200):
@@ -839,7 +839,7 @@ def checkTurn(iGameTurn, iPlayer):
 		
 		# third goal: control 9% of the world's population in 1940 AD
 		if iGameTurn == getTurnForYear(1940):
-			if calculatePopulationPercent(iIndonesia) >= 9.0:
+			if getPopulationPercent(iIndonesia) >= 9.0:
 				win(iIndonesia, 2)
 			else:
 				lose(iIndonesia, 2)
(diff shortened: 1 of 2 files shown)


Assets/Python/Victory.py
```diff
@@ -2811,6 +2811,16 @@ def countFirstDiscovered(iPlayer, iEra):
 			iCount += 1
 	return iCount
 	
+def isFirstDiscoveredPossible(iPlayer, iEra, iRequired):
```
dguenms, line 2877: Wouldn't be better to just count techs first discovered by `iPlayer` and all undiscovered techs `countFirstDiscovered(-1, iEra)` and see if they add up to at least `iRequired`?

dguenms (changes requested):

Assets/Python/Victory.py
MeuhV1: Yeah. I'll change it.

MeuhV1:
Updated

Assets/Python/Victory.py
```diff
@@ -1599,18 +1599,24 @@ def onTechAcquired(iPlayer, iTech):
...
 				else: lose(iJapan, 2)
+			if not isFirstDiscoveredPossible(iJapan iGlobal, 8) or not isFirstDiscoveredPossible(iJapan, iDigital, 8):
+				lose(iJapan, 2)
 				
 		# third English goal: be the first to discover ten Renaissance and ten Industrial technologies
 		if isPossible(iEngland, 2):
 			if countFirstDiscovered(iPlayer, iRenaissance) >= 8 and countFirstDiscovered(iPlayer, iIndustrial) >= 8:
 				if iPlayer == iEngland: win(iEngland, 2)
 				else: lose(iEngland, 2)
+			if not isFirstDiscoveredPossible(iEngland iRenaissance, 8) or not isFirstDiscoveredPossible(iEngland, iIndustrial, 8):
```
dguenms: Looks like there is a comma missing here for all these invocations.

dguenms (changes requested):

MeuhV1:
Updated.

I also included an additional condition so the UHV checks only if the discovered tech belongs to the era of the tech goals for speed improvement.

MeuhV1:
Rebased to latest DoC
Included Tamil UHV update
Fixed Enter first era goals

Assets/Python/Victory.py
```diff
@@ -532,21 +532,21 @@ def checkTurn(iGameTurn, iPlayer):
 				
 		# third goal: acquire 4000 gold by trade by 1200 AD
 		if isPossible(iTamils, 2):
-			iTradeGold = 0
+			iTradeGold = 0.0
```
dguenms: Instead of a float, this should still store an integer but instead by a factor of 100, Civ4 does not need more accuracy.

dguenms (changes requested):

Assets/Python/Victory.py
MeuhV1: Only iTradeGold of all instances of fTamilTradeGold?

dguenms: All instances where floats are used to store the accrued Tamil gold.

MeuhV1: Done.

I also included Italy to the list of conditional spawns and moved the code inside the check of line 1129.

MeuhV1:
Used int * 100 for Tamil UHV

I also included Italy to the list of conditional spawns and moved the code inside the check of line 1129 of RNF.py

dguenms (approved):

PR #217, opened 2018-05-27
Map

MeuhV1:


dguenms (approved):
Recently I have also noticed that if you place a landmark that isn't associated with a player (i.e. with the world builder tool) and then save from WB you trigger an exception and the resulting file is in a corrupted state. I think it is because it is trying to create a player instance to write a comment into the file. It seems related to this, can you also look into it?

MeuhV1:
I'll look into it.

MeuhV1:
Got it to work. :D

Assets/Python/pyWB/CvWBDesc.py
```diff
@@ -1809,7 +1809,10 @@ def write(self, f, sign):
 		f.write("BeginSign\n")
 		f.write("\tplotX=%d\n" %(sign.getPlot().getX(),))
 		f.write("\tplotY=%d\n" %(sign.getPlot().getY(),))
-		f.write("\tplayerType=%d, (%s)\n" %(sign.getPlayerType(), gc.getPlayer(sign.getPlayerType()).getName().encode(fileencoding)))
+		if sign.getPlayerType() == -1:
```
dguenms: Are you sure this works? Where is iPlayer defined?

MeuhV1:
You're right about iPlayer not being defined. I found what caused the error and got it to work. After a successful test, I removed some testing code. Apperantly I removed 1 line too many. But this does fix it. (I tested it just before the push to be sure)

Assets/Python/pyWB/CvWBDesc.py
```diff
@@ -1809,7 +1809,11 @@ def write(self, f, sign):
 		f.write("BeginSign\n")
 		f.write("\tplotX=%d\n" %(sign.getPlot().getX(),))
 		f.write("\tplotY=%d\n" %(sign.getPlot().getY(),))
-		f.write("\tplayerType=%d, (%s)\n" %(sign.getPlayerType(), gc.getPlayer(sign.getPlayerType()).getName().encode(fileencoding)))
+		iPlayer = sign.getPlayerType()
+		if sign.getPlayerType() == -1:
```
dguenms: This can now also be `iPlayer`

Assets/Python/pyWB/CvWBDesc.py
```diff
@@ -1809,7 +1809,11 @@ def write(self, f, sign):
 		f.write("BeginSign\n")
 		f.write("\tplotX=%d\n" %(sign.getPlot().getX(),))
 		f.write("\tplotY=%d\n" %(sign.getPlot().getY(),))
-		f.write("\tplayerType=%d, (%s)\n" %(sign.getPlayerType(), gc.getPlayer(sign.getPlayerType()).getName().encode(fileencoding)))
+		iPlayer = sign.getPlayerType()
```
dguenms, line 1812: This looks weird on Github, as if there was an additional linebreak. Are there trailing whitespaces?

dguenms (changes requested):

Assets/Python/pyWB/CvWBDesc.py
MeuhV1: The is no additional linebreak. There are also no trailing whitespaces. I guess it is just Github being weird.

I posted the new code in the post in the link below, so you can view it without Github peculiarities.
https://forums.civfanatics.com/threads/modmodding-q-a-thread.516755/page-94#post-15139548

EDIT:
Reuploaded to include iPlayer == -1, as you suggested.

dguenms (approved):

PR #218, opened 2018-06-23
Wonder art

MeuhV1:
New Angkor Wat art from RI
New Great Cothon art (included small change so it isn't in the coast as much)
New Harmandir Sahib art
New Kaaba art for Islamic Shrine
New Prambanan art (Temple complex using RFCA model)
Added Mole Antonelliana art (not assigned to a wonder yet)
New Khajuraho art (Akshardham)
Updated art of Channel Tunnel


dguenms:
The new Harmandir Sahib could be a bit larger. The Prambanan art is also very small, while the real wonder is very striking with the impressive size of its Stupas. The National College art is very unimpressive, in particular I would like a better button.

MeuhV1:
I didn't change the sizes of the new wonders in this PR, only the model and the skin. Locally I have an updated XML which have better wonders sizes of all wonders (both ingame and in the pedia). I didn't have that ready when I made the PR.

I agree that the National College art isn't the most astounding art, but I think it works well as a generic National College art. At least until some better art is found. I'm not pleased with the button either, but I didn't find a better one yet. I intended this button to be a placeholder.
I wanted to find some (temporary) art for the National College, so the old art could be used for the Oxford University.

dguenms:
We're in no hurry to replace the National College, especially not if the replacement is unsatisfactory.

Looking forward to the rescaled wonder art.

MeuhV1:
Updated the PR:
Rebased on the latest commit with new wonders
Reverted National College art
Small update in Channel Tunnel art
Resized wonders (and national wonders too)
Updated the Lsystem of some wonders
Fixed St. Basil Cathedral help text
Fixed a python exception in religions pedia

CvGameCoreDLL/CvTeam.cpp
```diff
@@ -4118,6 +4118,16 @@ void CvTeam::setDefensivePact(TeamTypes eIndex, bool bNewValue)
 				{
 					kPlayer.updateTradeRoutes();
 				}
+				// Merijn: Berlaymont: 1 free specialist per defensive pact
```
dguenms: Please don't include effect implementations here.

dguenms (changes requested):

CvGameCoreDLL/CvTeam.cpp
MeuhV1: Oops. I didn't intend to upload this at all. I forgot to deselect it. I will reupload it and exclude it.

I coded some new wonders effects myself. The sole reason I did it is that it was fun to to. I didn't even expect you to use my code but rather want to code it yourself. I don't consider my time wasted as I enjoyed it, which is all matters to me.

MeuhV1: Done

dguenms (approved):

PR #221, opened 2018-07-14
Traduction française 1

etiennefd:
Includes:
-Dawn of Man
-Religions
-Regions
-Buildings (but not the building strategy texts)
-Improvements


Assets/XML/Text/Regions.xml
```diff
@@ -12,303 +12,455 @@
 	<TEXT>
 		<Tag>TXT_KEY_REGION_0</Tag>
 		<English>Britain</English>
-		<French>Britain</French>
+		<French>
```
dguenms, line 15: The formatting in this file looks wrong.

dguenms (changes requested):

Assets/XML/Text/Regions.xml
etiennefd: Not sure why, apparently the kind of tab was wrong. Should be fixed now.

dguenms (approved):

PR #223, opened 2018-07-21
First part of German Translation

Sarius1997:
Translated the following files:

Actions.xml
Features.xml
Improvments.xml
Terrain.xml

Started with translating Dynamic Names


Assets/XML/Text/Resources.xml
```diff
@@ -95,7 +99,7 @@
 		</French>
 		<German>
 			<Text>Hitsingles</Text>
-			<Gender>Male</Gender>
+			<Gender>Female</Gender>
```
dguenms, line 102: Should also be plural, right?

PR #224, opened 2018-07-21
Traduction française 3

etiennefd:
Techs, civics, UPs, national wonders, various small fixes


Assets/XML/Text/Technologies.xml
```diff
@@ -170,24 +178,29 @@
...
+		<Italian>Le prime testimonianze sull'uso del bronzo, una lega di rame e stagno, risalgono alla Thailandia del 4500 a. C. Diversi secoli pi� tardi, la lavorazione del bronzo venne scoperta nel Medio Oriente. Gli oggetti risalenti a questo periodo dimostrano che il bronzo veniva inizialmente utilizzato solo a scopo decorativo, questo perch� lo stagno non era facilmente reperibile. Bisogner� aspettare pi� o meno il 2000 a. C. perch�, con l'importazione regolare di stagno dalle regioni confinanti, il bronzo sostituisca il rame nella fabbricazione di oggetti pi� grandi, offrendo anche una maggior resistenza rispetto al rame e la possibilit� di produrre punte pi� acuminate. Strumenti, armi e armature di bronzo erano anche meno costosi da realizzare e pi� duraturi. Il periodo di tempo in cui si preferiva il bronzo per tali prodotti � noto come l'Et� del Bronzo, che dur� fino al 1200 a. C. circa, quando fu soppiantato dalla lavorazione del ferro.</Italian>
```
dguenms: Can you run this through the converter too? The original text is to blame, but your editor likely made the problem worse.

Assets/XML/Text/Technologies.xml
```diff
@@ -382,15 +411,19 @@
...
+		</French>
 		<German>Geography</German>
 		<Italian>Geography</Italian>
 		<Spanish>Geography</Spanish>
 	</TEXT>
 	<TEXT>
 		<Tag>TXT_KEY_TECH_GEOGRAPHY_QUOTE</Tag>
 		<English>"Leave no stone unturned."[PARAGRAPH:2]- Euripides, Heraclidae</English>
-		<French>"Leave no stone unturned."[PARAGRAPH:2]- Euripides, Heraclidae</French>
+		<French>"La g&#233;ographie embrasse la totalit&#233; des choses, de m&#234;me que l'image d'une t&#234;te la repr&#233;sente tout enti&#232;re."[PARAGRAPH:2]- Ptol&#233;m&#233;e, G&#233;ographie</French>
```
dguenms: That's a good quote, but I would like to keep the actual quotes consistent between translations. Are the precedents of different quotes in the base game?

Maybe this one should feed back into the English text anyway.

Assets/XML/Text/Technologies.xml
```diff
@@ -293,15 +314,19 @@
...
+		</French>
 		<German>Artisanry</German>
 		<Italian>Artisanry</Italian>
 		<Spanish>Artisanry</Spanish>
 	</TEXT>
 	<TEXT>
 		<Tag>TXT_KEY_TECH_ARTISANRY_QUOTE</Tag>
 		<English>"The life so short, the craft so long in the learning."[PARAGRAPH:2]- Geoffrey Chaucer, Parliament of Foules</English>
-		<French>"The life so short, the craft so long in the learning."[PARAGRAPH:2]- Geoffrey Chaucer, Parliament of Foules</French>
+		<French>"Bien qu’on ait du c&#339;ur &#224; l’ouvrage, l’Art est long et le Temps est court."[PARAGRAPH:2]- Charles Baudelaire, Les Fleurs du mal</French>
```
dguenms: Same here.

Assets/XML/Text/Technologies.xml
```diff
@@ -676,15 +729,15 @@
 			<Plural>0</Plural>
 		</Italian>
 		<Spanish>
-			<Text>Cartograf�a</Text>
+			<Text>Cartograf�a</Text>
```
dguenms: Converter?

Assets/XML/Text/Technologies.xml
```diff
@@ -1684,15 +1785,19 @@
 	<TEXT>
 		<Tag>TXT_KEY_TECH_FILM</Tag>
 		<English>Film</English>
-		<French>Film</French>
+		<French>
+			<Text>Cin&#233;ma</Text>
```
dguenms: "Film" here is meant to reference the medium more than "moving pictures", and also include photo film etc. English has the advantage of covering both with one word but cinéma seems to be more in the direction of moving pictures.

Assets/XML/Text/Technologies.xml
```diff
@@ -2464,7 +2627,7 @@
...
+		<Italian>Le prime civilt� tribali erano legate tra loro da esigenze e credenze comuni. Questi gruppi erano guidati da capi trib� che agivano da consiglieri e da interpreti della volont� divina. In queste culture primitive non esistevano vere e proprie leggi, ma il timore degli dei e un senso di tradizione e morale tribale mantenevano l'ordine all'interno delle comunit�. Man mano che le societ� si sviluppavano e si differenziavano, sorse per tutte l'esigenza di stabilire delle regole di comportamento. I primi codici di leggi conosciuti furono redatti a Babilonia, in India e in Palestina. Le Dodici Tavole di Roma, scritte nel 500 a. C., e il successivo Codice di Giustiniano, furono i primi codici a distinguere tra diritto pubblico, che riguarda lo Stato, e diritto privato, che riguarda le dispute tra cittadini. Il diritto Romano fu il primo sistema scritto di leggi e continua a influenzare ancora oggi molti sistemi legali.</Italian>
```
dguenms: Converter?

Assets/XML/Text/Technologies.xml
```diff
@@ -4118,15 +4413,19 @@
 	<TEXT>
 		<Tag>TXT_KEY_TECH_STATECRAFT</Tag>
 		<English>Statecraft</English>
-		<French>Statecraft</French>
+		<French>
+			<Text>Statecraft</Text>
```
dguenms: Really?

dguenms (changes requested):

Assets/XML/Text/Technologies.xml
etiennefd: Didn't know what to translate it to, skipped it, and forgot about it. Will try to find something, it's hard in part because I'm not too sure what the tech is supposed to mean, besides more advanced administrative practices.

Assets/XML/Text/Technologies.xml
etiennefd: Sure, I'll do it.

Assets/XML/Text/Technologies.xml
etiennefd: In this case, I looked up the actual French translation of the Heraclidae, and found that it didn't work at all—the idiom is translated into something different, as idioms tend to be, and unrelated to geography. So I tried to find something else but also ancient Greek. Let me know what you want to do in this case (I can change the English to Ptolemy too, to be honest I'm not crazy about the Euripides quote).

Assets/XML/Text/Technologies.xml
etiennefd: Here, I couldn't find a French translation of the Parliament of Foules, and then I noticed that the "Ars longa, vita brevis" aphorism had also been used by Baudelaire, a major French poet, so it seemed appropriate to use that instead. 

I agree that quotes should be consistent between languages in general, but I'd say it's fine if there are a couple of exceptions when it feels right. I believe there was one instance of a quote that didn't match the English one when I used to play Civ 4 in French. 

Let me know what you prefer. (If you'd like to use the Baudelaire quote for English, there are like 4 different translations [here](https://fleursdumal.org/poem/110).)

Assets/XML/Text/Technologies.xml
etiennefd: Makes sense, I'll change it back

Assets/XML/Text/Technologies.xml
etiennefd: Maybe "Sciences de l'administration"? Or "Sciences politiques"? Or simply "État"?

Assets/XML/Text/Technologies.xml
dguenms: Yes, let's use the Ptolemy quote in English too.

Assets/XML/Text/Technologies.xml
dguenms: Okay, I think it makes sense here since they are basically different takes on the original Latin phrase.

etiennefd:
All comments are addressed.

dguenms (approved):

PR #227, opened 2018-08-09
Art for Crystal Cathedral, Tokyo Skytree and Las Lajas Sanctuary

MeuhV1:
All with XML provided. The models have the correct ingame and pedia sizes.

I gave Las Lajas Sanctuary the LRIVER LSystem, which places the bridge of the model over the water. Like a bridge. If the city does not borders a river, it will be placed in the center of a city. But maybe you want to add the river requirement. (I don't know on top of my head if the new map has the Guaitara river present) I leave that up to you.


MeuhV1:
2nd commit:
Updated my other models with intended colors
Reduced the size of the nif files of several wonders and buildings

PR #233, opened 2018-08-26
Chateau Frontenace art

MeuhV1:
You didn't give a definite confirmation you would include it, but you said you liked it. If you do intend to include it, here is the art. Otherwise just reject the PR.


dguenms:
Thanks. I am not sure about the potential effect yet but I'll let myself be inspired by the art. In the meantime I will wait on merging this.

dguenms:
Sorry, I forgot merging this before adding new art, can you resolve the conflict?

PR #236, opened 2018-09-18
Fix DOM text to work with localization

marajxa:
### Testing
Loaded up several civs, in different scenarios, in English and in French, and checked that text rendered correctly.


dguenms (changes requested):
As I mentioned in the thread, I prefer a solution where the identifier is identical to the short description, and the text keys currently using the identifier are changed.

marajxa:
Apologies, got hit by the fall holidays and then sickness in quick succession. Having some git problems, but hopefully I will get this up soon.

dguenms:
Hey, no worries. As you may have noticed I didn't have a lot of time for DoC as well lately. I hope you feel better now.

PR #238, opened 2018-09-27
Improved Great People 7

etiennefd:
Been accumulating Great People name for a while. Two changes you may or may not want: 
- I reordered the list by civ spawn date
- I merged Greece and Byzantium: Byzantine GP now make up the medieval era for Greece. It makes sense since Byzantium was in many ways culturally continuous with Greece and now if you play Greece in the medieval era you'll get era-appropriate names (and similarly for playing Byzantium into the industrial/global eras)


Assets/Python/GreatPeople.py
```diff
@@ -3504,14 +3518,18 @@ def setup():
 		"Clemens Brenninkmeijer", # 19th
 		"August Kessler", # 19th
 		iGlobal,
+		"Anton Philips", # 20th
```
MeuhV1: There is also a Frits Philips in the Great Engineers list. (Which is the son of Anton) I think it's better to only use one of them. (Anton in probably better then Frits) And I think they fit the role of Great Engineer more than Great Merchant.

etiennefd: Thanks for verifying the Dutch GP! I'll change that.

dguenms:
I am not a fan of the merged Greek/Byzantine list because 3000 BC Byzantium does not start in the medieval era, adding a lot of classical Greek GP names which would feel inappropriate for them. I see the desire for letting them use a merged list from the medieval era on but this should be solved differently.

dguenms:
I'm agnostic about reordering the GP lists, but please include them in a different pull request. I can easily approve a PR that does nothing but reordering things but when combining this with additional other changes it is near impossible for me to see from the diff what they are.

My suggestion on how to go from here:
1. You make a new PR that only includes the reordering, I'll quickly merge it
2. You make a second PR that includes all other changes besides the Greek/Byzantine merge so I can review them
3. We figure out a way to partially share lists between civs, e.g. by having the list selection also be era dependent

etiennefd:
Sounds good. 

I understand and share your concern for the Byzantines. Making the list era-dependent and creating a Byzantine list for the Classical era only sounds like it would work elegantly.

dguenms:
Okay, let's go with that then. Thanks for all the effort you put into this.

PR #246, opened 2018-10-05
Fixed two python errors

MeuhV1:
- Top Cities when there are less then 5 cities in the world
- Removed Atomium part of RFCEventHandler. The implementation is in CvCity::processBuilding() and CvPlayer::changeNumNukeUnits(). (If not, sd.iNumNuclearWeapons should be changed to data.iNumNuclearWeapons and both iNumNuclearWeapons and tAtomiumPlot should be added to StoredData.py)

### DIFF
--- a/Assets/Python/CvRFCEventHandler.py
+++ b/Assets/Python/CvRFCEventHandler.py
@@ -612,11 +612,6 @@ class CvRFCEventHandler:
 				if plot.getOwner() == iOwner and not plot.isWater():
 					plot.setWithinGreatWall(True)
 					
-		# Leoreth: Atomium
-		if iBuildingType == iAtomium:
-			data.tAtomiumPlot = (city.getX(), city.getY())
-			city.changeBuildingCommerceChange(gc.getBuildingInfo(iAtomium).getBuildingClassType(), CommerceTypes.COMMERCE_RESEARCH, sd.iNumNuclearWeapons)
-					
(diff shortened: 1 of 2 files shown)


Assets/Python/Screens/CvInfoScreen.py
```diff
@@ -1773,7 +1773,7 @@ def calculateTopCities(self):
 					+ city.getYieldRate(YieldTypes.YIELD_COMMERCE))) * city.getPopulation()
 				lCities.append((city, iValue))
 		lCities.sort(key=itemgetter(1), reverse=True)
-		for i in range(5):
+		for i in range(min(len(lCities), 5)):
```
dguenms: I'd prefer
```
for i, city in enumerate(lCities):
```
here, followed by using the `city` object directly.

Assets/Python/CvRFCEventHandler.py
```diff
@@ -612,11 +612,6 @@ def onBuildingBuilt(self, argsList):
 				if plot.getOwner() == iOwner and not plot.isWater():
 					plot.setWithinGreatWall(True)
 					
-		# Leoreth: Atomium
```
dguenms, line 615: True, I'm confused now why this is left over.

dguenms (changes requested):

MeuhV1:
Done

dguenms (approved):

PR #249, opened 2018-10-08
Revise event texts

Alexius08:
- Update quest texts to reflect changes on tech tree and unit roster (in
English, and, whenever available, in German, using the provided
translations)
- Copy translations for Harbormaster quest from old BTS text (needs to
be revised)
- Remove duplicate entries

### DIFF
--- a/Assets/XML/Text/Events.xml
+++ b/Assets/XML/Text/Events.xml
@@ -283,11 +283,11 @@
 	</TEXT>
 	<TEXT>
 		<Tag>TXT_KEY_EVENT_BEST_DEFENSE_HELP_1</Tag>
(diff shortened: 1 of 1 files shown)


Assets/XML/Text/Events.xml
```diff
@@ -1596,15 +1596,15 @@
 	</TEXT>
 	<TEXT>
 		<Tag>TXT_KEY_EVENT_GUNS_BUTTER_DONE_3</Tag>
-		<English>Enjoy our Golden Age of Muskets.</English>
+		<English>Enjoy our Golden Age of Arquebuses.</English>
```
dguenms: Maybe it should be something like the Golden Age of Gunpowder?

dguenms (commented):
Looks good. Due to another change you have to rebase on develop again, unfortunately.

Also, if you changed any game text entry, can you also remove the text for any language you did not adjust accordingly and replace it with the English one? Otherwise it's impossible to find out which texts require partial translations.

dguenms:
Also, is this changeset exhaustive in your opinion?

Alexius08:
Yes. Only the following events needed to have their texts updated:

- Guns, Not Butter
- Overwhelm
- Noble Knights
- Reinforced Hull
- Best Defense

PR #256, opened 2018-12-30
City screen wonder limit indicator

MeuhV1:


dguenms:
Can you split the tooltip so there is one each for national and world wonders?

MeuhV1:
Sure.

MeuhV1:
Done. Also rebased to the latest version.

PR #262, opened 2019-01-12
China update

MeuhV1:


dguenms (commented):
Those techs were actually only intended for the AI, but thanks for catching that. I have fixed it myself.

PR #264, opened 2019-02-03
Improved Great People 8

etiennefd:
The main change is that I split Holy Rome and Germany, and added many names to both. The split is from 18th century onwards, so all pre-1700 Germans are in the Holy Rome list, as are post-1700 people who are more closely associated with Austria (including a few Hungarians and Czechs). Post-1700 people who are more closely associated with Prussia or modern Germany are in the Germany list. 

Things are not always clear cut for 18th c. people who were not Prussian or Austrian (Schiller? Goethe?). One possibility would be to make the split only from the industrial era onwards. Let me know. I don't claim any sort of particular knowledge of German culture and history, so I might have made some questionable choices. One thing I'm sure of is that Gutenberg and Kafka really needed to be added to the list. 

Other changes:
- Ethiopia: removed Ezana and Menelik, added Haile Selassie as a prophet.
- Added Publicola's suggestions for Mongolian GPs.
- Moved the Maya and China to preserve spawn order.
- Various other additions and corrections.


dguenms:
Haven't looked into the changes in depth, but a possible solution for the neither Prussian nor Austrian GPs is to just include them in both lists. The game will still consider an already used name taken and not spawn them for the other civ.

etiennefd:
That's probably a good idea. I'll go over the lists and change them accordingly.

Assets/Python/GreatPeople.py
```diff
@@ -3263,13 +3312,19 @@ def setup():
 		"Dayan Khan", # 15th
 		iIndustrial,
 		"Balingiin Tserendorj", # 19th
+		iGlobal, 
+		"Damdin Sükhbaatar", # 20th
```
dguenms: u""

Assets/Python/GreatPeople.py
```diff
@@ -2493,68 +2515,76 @@ def setup():
...
-		"Paul Klee", # 20th
-		"fLeni Riefenstahl", # 20th
-		"Leoreth", # 21st
+		"Egon Schiele", # 20th
+		"Fritz Lang", # 20th
 	],
 	iGreatScientist : [
-		"Nikolaus von Kues", # 15th
+		"Hrabanus Maurus", # 9th
+		"fHildegard von Bingen", # 12th
```
dguenms: Hildegard von Bingen should stay a prophet imo

dguenms (changes requested):

dguenms (approved):

PR #265, opened 2019-03-08
New-World Music Module

civplayer:
This is the first installment, of a proposed expansion of music of DOC to cover musically underrepresented areas and to expand on an already amazing Soundtrack. This particular branch consists of 'New World' or Precolombian music. Music represents the indigenous cultures of North, Central and South America. The majority of the United States- Indian music comes from Civ IV Colonization and Carlos Nakai. However much of the Aztec music comes from a unique album from Ricardo Lozano that is the most accurate portrayal of Mesoamerican music to the best of my knowledge. To make this possible I have coded the creation of a New-World musical era so that the music should play for the Maya, Inca and Aztec between the classical and Rennaissance eras. Please give this a try and give feedback!


civplayer:
Is the request coming through?

MeuhV1:
I can see it.

dguenms:
You have just added a bunch of new files. But I will take a look at the music files.

civplayer:
The other files are the audio 2dscripts, the audio defines and the dll file which are to make the soundtrack functional

I will say in advance though is that I think I handled the dll file incorrectly I didn't realize that when I uploaded.

Please review the music  :). I tried to make it both authentic and aesthetically pleasing.

dguenms:
It's just that you created new files in their entirety instead of editing the existing files.

But I don't think the code changes are the relevant part of this proposal. I'll see if I can make use of the music included here.

civplayer:
I thought making the new files would make sure I was not corrupting any existing file.
What do you think of these tracks? Are there specific things you would like removed? Or, something in particular else you would like to see here. I tried to avoid too many 'voiced songs' though in this regional genre that can be a bit of a challenge.

Also I would also suggest (this might be in another pull request) adding one or two of these tracks to the European Renaissance category, as every civ in the game presented (except Poland and Turkey) were somehow involved in the Colombian exchange.

civplayer:
Thanks for considering! :)
I am downloaded Window Visual studios to properly edit files, but before that would it be okay if I also began to upload some of my other music proposals?

dguenms:
Whatever works best for you. It will not be before the release of 1.16 that I will have attention for music related changes.

PR #272, opened 2019-04-16
Feature - Building Bonus Yield AI

1SDANi:
I tried to implement AI for the bonus yield building effect. I tried to emulate existing code for it.

EDIT: Seems I accidentally included a commit that occurred when I pulled the latest develop.

### DIFF
--- a/CvGameCoreDLL/CvCityAI.cpp
+++ b/CvGameCoreDLL/CvCityAI.cpp
@@ -4056,7 +4056,7 @@ int CvCityAI::AI_buildingValueThreshold(BuildingTypes eBuilding, int iFocusFlags
 						// 1SDAN
 						if (kBuilding.getBonusYieldChange(iJ, iI) > 0)
 						{
-							iTempValue += (kBuilding.getBonusYieldChange(iJ, iI) * countNumBonusPlots() * getBaseYieldRate((YieldTypes)iI) / 6);
+							iTempValue += (kBuilding.getBonusYieldChange(iJ, iI) * countNumBonusPlots() * getBaseYieldRate((YieldTypes)iI) * 6);
 						}
 					}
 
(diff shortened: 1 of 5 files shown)


CvGameCoreDLL/CvCityAI.cpp
```diff
@@ -4052,6 +4052,12 @@ int CvCityAI::AI_buildingValueThreshold(BuildingTypes eBuilding, int iFocusFlags
 						{
 							iTempValue += ((kBuilding.getBonusYieldModifier(iJ, iI) * getBaseYieldRate((YieldTypes)iI)) / 12);
 						}
+
+						// 1SDAN
+						if (kBuilding.getBonusYieldChange(iJ, iI) > 0)
+						{
+							iTempValue += (kBuilding.getBonusYieldChange(iJ, iI) * countNumBonusPlots() * getBaseYieldRate((YieldTypes)iI) / 6);
```
dguenms: I think this values the yield change too lowly. Compare above `kBuilding.getYieldChange(iI) * 6` for the flat yield increase from a building.

dguenms (changes requested):
Besides that it looks good. Can you remove the DLL itself from the pull request? This way I can merge this later while committing further DLL changes in the meantime.

CvGameCoreDLL/CvCityAI.cpp
1SDANi: I'm gonna be honest, I didn't know what to put there so I just kind of guessed based on the nearby values. What do you think I should put?

Also, how do I remove the DLL from the pull request?

dguenms: Yeah, I also never know what to use so I just try to keep it consistent with what's already there. I see no reason not to weigh yield from improvements the same way as yields from buildings, so `*6` should be the right factor.

If you're using Tortoise, it has a "delete (keep local)" option that should do the right thing.

PR #277, opened 2019-04-26
Feat building bonus yield ai

1SDANi:
I accidentally did another stupid. Attempt number 3

### DIFF
--- a/CvGameCoreDLL/CvCityAI.cpp
+++ b/CvGameCoreDLL/CvCityAI.cpp
@@ -4056,7 +4056,7 @@ int CvCityAI::AI_buildingValueThreshold(BuildingTypes eBuilding, int iFocusFlags
 						// 1SDAN
 						if (kBuilding.getBonusYieldChange(iJ, iI) > 0)
 						{
-							iTempValue += (kBuilding.getBonusYieldChange(iJ, iI) * countNumBonusPlots() * getBaseYieldRate((YieldTypes)iI) / 6);
+							iTempValue += (kBuilding.getBonusYieldChange(iJ, iI) * countNumBonusPlots() * getBaseYieldRate((YieldTypes)iI) * 6);
 						}
 					}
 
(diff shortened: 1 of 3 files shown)


1SDANi:
Why aren't the latest 2 commits included???

1SDANi:
Hmm, it seems that it was switched to * 6 despite the commit not being there, but the removal of the DLL didn't work. 

EDIT: It says -5.2 MB. Did I just commit the removal of the DLL? I guess it did work after all.

CvGameCoreDLL/CvCityAI.cpp
```diff
@@ -4052,6 +4052,12 @@ int CvCityAI::AI_buildingValueThreshold(BuildingTypes eBuilding, int iFocusFlags
 						{
 							iTempValue += ((kBuilding.getBonusYieldModifier(iJ, iI) * getBaseYieldRate((YieldTypes)iI)) / 12);
 						}
+
+						// 1SDAN
+						if (kBuilding.getBonusYieldChange(iJ, iI) > 0)
+						{
+							iTempValue += (kBuilding.getBonusYieldChange(iJ, iI) * countNumBonusPlots() * getBaseYieldRate((YieldTypes)iI) * 6);
```
dguenms: Why is it multiplied by the base yield rate? That's definitely not right.

dguenms (commented):
I think now the PR removes the entire DLL from the repository.

The best solution probably is to undo all your commits (which keeps the file changes, turning them into uncommitted changes again), and then commit everything except the DLL again and force push the new commit.

[This here](https://stackoverflow.com/a/22657460) should explain how to do it.

CvGameCoreDLL/CvCityAI.cpp
1SDANi: I may or may not have thought that represented the value of the yield to the AI. Fix incoming.

dguenms: It's actually the total yield produced by the city. It shows up in a number of other building effect evaluations because they are a percentage change that affects the total.

1SDANi: For some reason I didn't notice the top level comment before. Will do.

PR #286, opened 2019-05-01
Korean DynCityName update LITE

mechu1205:
Features
Korean translations for various Chinese/Manchurian & every Japanese city names.
Incorporated Jilin, Jillin and Chilin into one single index.
Korean cities in Manchuria change name to appropriate Chinese names when captured.
Fixed Chinese names map to identify Pusan (previously the tile was wrongly identified as Hanseong/Seoul)
No other changes have been made to the names lists and maps.


Assets/Python/CityNameManager.py
```diff
@@ -5626,14 +5703,82 @@ def onRevolution(iCiv):
 #Language: Korean
 {
 	"Beijing"		:	"Bukgyeong",
-	"Zhongdu"		:	"Bukgyeong",
+	"Zhongdu"		:	"Bukpyeong",
```
dguenms, line 5706: There is still a typo here.

dguenms (changes requested):

Assets/Python/CityNameManager.py
mechu1205: that's not a typo, that's the korean pronunciation of Beiping, the name of Beijing in the Ming era.
Since Beijing has two separate names that change through eras in-game,
I separated the Korean names as well.

dguenms: But apparently Bukpyeong does not map to Zhongdu? I don't think it makes sense to handle names this way. It would be different if Beiping existed as a name but a name that does not exist for the native language should not be reflected in another language.

mechu1205: Ok, I'll revert Bukpyeong to Beijing.

dguenms (approved):

PR #301, opened 2019-06-27
German translation for War Chariot Strategy and Desert Adaptation Promotion

Imperator-Knoedel:


dguenms:
Why are so many commits in this PR? It should only be the last one.

Assets/XML/Text/Promotions.xml
```diff
@@ -105,7 +105,7 @@
 			<Gender>Female</Gender>
 			<Plural>0</Plural>
 		</French>
-		<German>Desert Adaptation</German>
+		<German>W&#252;sten Anpassung</German>
```
dguenms: That's a deppenleerzeichen. Please use Wüsten-Anpassung or Wüstenanpassung

dguenms (changes requested):

Imperator-Knoedel:
> Why are so many commits in this PR? It should only be the last one.

That's a good question, I'll look into it.

PR #302, opened 2019-06-28
Forests varieties

MeuhV1:
Bamboo forest
Hybrid forest
Feature variety button


dguenms (approved):
As I said I want to give it a look in the game as well but from a code perspective it looks good.

PR #304, opened 2019-07-22
Various updates

MeuhV1:
Corporations
- Global resource cap for corporations instead of individual resource cap
- Fixed Brazilian UP not using the number of Sugar resources
- Fixed Brazilian UP not updating commerce from corporations when the amount of available sugar resources changes
- Tooltip uses CvCity::getCorporationYieldByCorporation() and CvCity::getCorporationCommerceByCorporation() to calculate bonus instead of own (wrong) calculations
- Tooltip for Brazil shows sugar as consumable resource for Oil Industry

Power consumption
- BUG actualBuildingHelp takes amount of buildings that consume power in account when building power buildings
- BUG actualBuildingHelp now includes building that consume (dirty) power when calculating actual health

Financial advisor
- Added scroll function for income panel

Misc
- Removed comma in front of the button path in CivArtDefines.xml (It could lead to a CTD if this path is used by certain functions)
- Removed redundant XML files from the art folder


CvGameCoreDLL/CvGameTextMgr.cpp
```diff
@@ -14301,17 +14301,36 @@ void CvGameTextMgr::setCorporationHelpCity(CvWStringBuffer &szBuffer, Corporatio
 		}
 	}
 
-	int iNumResources = 0;
+	// int iNumResources = 0;
```
dguenms, line 14304: Why is this adding commented code?

MeuhV1: I replaced this with the bResources section. (lines 14314-14331 of the new code) iNumResources was only use once (line 14333 of the new code) and now uses the new bResources. It doesn't change the functionality, but it was easier to include the Brazilian UP and the global resource cap this way. I didn't remove the old code and only commented it out.

PR #306, opened 2019-07-30
Feature varieties

MeuhV1:
Commit 1:
Improved the savanna model files. The old files were the jungle model, but with a all bushes made transparent so only the trees were left visible. I removed all vertices that were not related to the trees or their shadows. I also fixed some shadows.

Commit 2:
Bamboo forest uses same broadleaf texture from Terrain2.fpk

Commit 3:
Tropical forest variety


dguenms:
See my comment in the Alternative Map development thread. I think we need to figure out the cause of some visual issues with the new feature textures.

MeuhV1:
I gave the texture files mipmaps. I don't know if this fixes the problems, but we can at least try.

dguenms:
That did it! Merge coming.

PR #307, opened 2019-08-02
Fixed yield display when having more than 10 yields

MeuhV1:


CvGameCoreDLL/CvPlot.cpp
```diff
@@ -1067,23 +1067,44 @@ void CvPlot::updateSymbols()
...
+		// merijn: 5x symbols
+		if (maxYield >= 5)
 		{
 			pSymbol = addSymbol();
 			for (int iYieldType = 0; iYieldType < NUM_YIELD_TYPES; iYieldType++)
 			{
-				int iYield = yieldAmounts[iYieldType] - (maxYieldStack * i);
-				LIMIT_RANGE(0,iYield, maxYieldStack);
-				if(yieldAmounts[iYieldType])
+				if (yieldAmounts[iYieldType])
```
dguenms: Can we make this an explicit boolean, i.e.
```
if (yieldAmounts[iYieldType] > 0)
```

Assets/XML/GlobalDefines.xml
```diff
@@ -187,7 +187,7 @@
 	</Define>
 	<Define>
 		<DefineName>MAX_YIELD_STACK</DefineName>
-		<iDefineIntVal>5</iDefineIntVal>
+		<iDefineIntVal>10</iDefineIntVal>
```
dguenms, line 190: What does this affect?

dguenms (changes requested):

CvGameCoreDLL/CvPlot.cpp
MeuhV1: No problem.

Assets/XML/GlobalDefines.xml
MeuhV1: It assigns 10 "slots" to each yield instead of the old 5. The new dds file consists of 32 slots of 64x64 images.

MeuhV1:
Updated

CvGameCoreDLL/CvPlot.cpp
MeuhV1: Done

dguenms (approved):

PR #308, opened 2019-08-03
Feature - Stability Change Notification

1SDANi:
Added a non-recorded notification when the player's stability level changes.


dguenms:
Thanks, but I don't think this is necessary.

PR #310, opened 2019-08-05
mipmaps

MeuhV1:


dguenms (changes requested):
As per Xyth's suggestion, it's not necessary to include mipmaps in interface elements, so symbols.dds and a couple of buttons included here should not change.

MeuhV1:
Reverted the buttons and yield symbols

dguenms (approved):

PR #311, opened 2019-08-08
Map

MeuhV1:
Commit 1:
Added missing turquoise button
Hide bonus varieties in pedia
Hide bonus varieties in WB bonus mode

Commit 2:
Added bonus variety mode
- Selectable list only contains variety bonuses (or do you want every resource in this list?)
- Left click to set bonus variety type. This only works on tiles with a resource.
- Right click to remove bonus variety. (and therefore set the bonus to default graphics)

Commit 3:
Fixed vanilla art

Commit 4:
Moved bonusVariety on non-resource tiles to DLL

Commit 5:
Fixed variety bonus-improvement interaction


Assets/Python/Screens/PlatyBuilder/CvPlatyBuilderScreen.py
```diff
@@ -514,6 +514,9 @@ def placeObject( self, bMulti=False ) :
 			self.m_pCurrentPlot.setImprovementType(self.iSelection)
 		elif self.iPlayerAddMode == "Bonus":
 			self.m_pCurrentPlot.setBonusType(self.iSelection)
+		elif self.iPlayerAddMode == "BonusVariety":
+			if self.m_pCurrentPlot.getBonusType(-1) != -1:
```
dguenms: I'm wondering if it might be a better idea to limit this on the DLL end already.

MeuhV1: Done.

dguenms (approved):
Great!

PR #315, opened 2019-09-03
Fixed murex waves

MeuhV1:


dguenms:
Looks good! Have you also tried to partially submerge the shells in water, like clam shells are?

MeuhV1:
Not yet. But that should be easily doable by simply moving the whole mesh in the Z-axis.

During testing of the wave animation, I noticed that the clams also move up and down themselves a bit. I can try to include that as well, but I’m not sure if I will be succesful in that.

dguenms:
Okay, let me know if you're done experimenting and have implemented everything that worked.

MeuhV1:
I updated the the model by lowering the mesh in the Z-axis. You can't really see the underside of the shells because of the convex shape of the shells. (This was also the case before I lowered it) It is a bit hard to see the submerged parts.

I was not able to include the movement. To fix the waves, I copied the murex mesh to the clams mesh. (This was easier than copying the waves to the murex model) I hoped that as a side effect the properties that move the clams would also apply to the new murex mesh, but that didn't happen unfortunately. I don't have much experience with animations and I currently have no idea how I should link the animation with the new mesh.

Maybe I will return to this in the future when I have better knowledge about animations.

MeuhV1:
I included a fix for the all region overlay in the WB
I fixed 2 tiles with the wrong region assigned (tile in Ethiopia: Brittain -> Ethiopia, tile in Alaska: Siberia -> American Arctic)

MeuhV1:
I rebased the branch to include the new csv map changes. I updated the region exporter to directly update the csv.

PR #319, opened 2020-02-23
New marsh variety and texture updates

MeuhV1:


dguenms:
Two things:
1. I liked the different textures for the new forest varieties, I think they were better before. I do prefer the Savanna texture introduced here though.
2. I see zero difference between the marsh varieties, both for the three currently existing varieties and the new one you introduced. They all look like the default variety to me. Have you tried with Blue Marble? Maybe it has issues.

dguenms:
Actually, never mind. I checked again and I also prefer Savanna unchanged.

MeuhV1:
1. Okay. I will revert the textures from the new forest varieties.

2. I do have blue marble installed. But I have no clue why they show up as the default variety for you.

dguenms:
That's strange, although it explains why I didn't notice the other marsh varieties earlier. I can select all the varieties, but when I apply them they always come out looking like the default variety. It makes no difference if the tile already had the marsh feature or not. I haven't checked if it's just a graphical error or if the attribute is not applied to the plot at all. I also haven't tried saving and reloading to force redrawing the map.

I won't have time to look deeper into this this week and unfortunately I'm not comfortable merging this until it shows up on my end.

MeuhV1:
Does it work when you remove the old marsh first and then place a new one with a different variety? I also can't change the variety by placing a new marsh feature directly. (Strangely, it does work with forests) But when I remove the existing one and replaced it with a new one, I do get the new variety.

MeuhV1:
When I try to change the variety of marshes, it does not work directly either. I have to remove the old marsh first and then replace the marsh with the selected variety for it to work. But this only applies if the tile already has a marsh. (Strangely I can change the variety of forests without needing to remove the old one first)

dguenms:
I only tried placing a new marsh variety on a tile that never had a feature. Are you saying that if it you remove a marsh feature and place a different variety on the tile it works?

MeuhV1:
Yes, but that is only required if that tile already has a marsh feature. If the tile has no feature or a different feature then it works properly.

dguenms:
True, it behaves like you describe. That's odd, but as long as I can place the varieties on the map I think it's alright.

PR #325, opened 2020-04-14
Rewrite

MeuhV1:
I rewrote the export functions of the WB. This is basically the same as point 5 of your "plans for 1.17". I also immediately applied the new functions of the core module. I removed some redundant functions (like checking if you made any changes before exporting). I temporarily disabled the function to export the maps of all civs at once, but this one will return in a future PR.

The rewrite code caused the stability overlay not to work properly anymore. I fixed the issues and in the process I made some minor code refactorizations and applied the new functions of the Core, like plots.all() etc. where I could. I also applied these similar changes to the DoC map view overlays().

I moved the debug() and show() functions from the Core back to RFCUtils. I couldn't make it work when I kept it in the Core. My suspicions is some kind of "import loop", which causes one of the files to not load properly or something. (Core import Popup, which imports CvUtil, which imports Core etc.) Moving it back to RFCUtils solved the problem. Although this might be against what you aim with the Core module.


dguenms:
Sorry for not acting on this sooner, but I was (and still am) addressing issues with the Core module and expanding it some more. I actually also ran into the issue with show(), and I figured WB was broken because I wasn't really checking if it survived the refactoring.

This PR is already in conflict since the rewrite branch moved on. I suggest you wait a bit until Core is a bit more stable before getting this ready to merge. I will notify you here.

MeuhV1:
Okay. That's fine.

One of the reasons for this PR is that I also noticed that the WB was broken a bit. (Actually the WB itself was not really broken. Only the overlay system had some errors) Prior to the core module was introduced I already had some ideas on how to improve the WB. I combined them into one PR for practical reasons.

In the meantime, I made some more updates to the WB, but didn't add them to the PR yet. I will include them in this PR when it the rewrite branch is ready to be merged with this PR. This PR should also fix the things that were broken in the WB because of the code changes.

dguenms:
Hey, even though it says so I didn't intentionally close this pull request. I think it was automatically closed because I deleted the rewrite branch this is pointing to. You are welcome to open the PR again for develop.

PR #327, opened 2020-05-26
fixed kelebolo not being able to enter jungle, rainforest, marsh

redaloe:


dguenms:
What is the issue you see here? Skirmishers generally are not able to enter these features.

redaloe:
Sorry for not being clearer. The Civilopedia entry for the Kelebolo says that it can enter these features, and it also has a bonus to rainforest attack, which wouldn't make sense if it couldn't enter rainforests in the first place, so when I noticed that it can't actually do that in game, I assumed that it was a bug.

dguenms:
Interesting that it says that. I will verify what's going on.

PR #328, opened 2020-06-03
Olmecs

etiennefd:
There's a bug due to a wrong ordering of civs in Consts.py: the Olmecs (instead of the Greek) can declare war on spawn on Babylonia


MeuhV1:
You selected the wrong base remote. Make sure you select 1SDAN's remote. ;)

civplayer:
If I wanted the Olmecs in the current version of DOC, how would I go about merging the files without causing conflicts?

etiennefd:
I suspect that this would be difficult since Leoreth changed a lot of the code around civilizations and scenarios in the recent updates. I think your best bet would be to "merge" manually, i.e. take all the Olmec-specific content from here and apply it where necessary to the updated code base.

PR #339, opened 2020-07-24
Yield updates when entering new era

MeuhV1:
This fixes the Moors yields issue.

### DIFF
--- a/CvGameCoreDLL/CvPlayer.cpp
+++ b/CvGameCoreDLL/CvPlayer.cpp
@@ -12203,6 +12203,12 @@ void CvPlayer::setCurrentEra(EraTypes eNewValue)
 			gDLL->getInterfaceIFace()->setDirty(Soundtrack_DIRTY_BIT, true);
 		}
 
+		//Merijn: Moorish UP: update the yield when UP ends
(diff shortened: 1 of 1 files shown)


CvGameCoreDLL/CvTeam.cpp
```diff
@@ -5652,6 +5652,9 @@ void CvTeam::setHasTech(TechTypes eIndex, bool bNewValue, PlayerTypes ePlayer, b
 					if (GET_PLAYER((PlayerTypes)iI).getCurrentEra() < GC.getTechInfo(eIndex).getEra())
 					{
 						GET_PLAYER((PlayerTypes)iI).setCurrentEra((EraTypes)(GC.getTechInfo(eIndex).getEra()));
+
+						// Merijn: Update yields when entering new era (for Moorish UP)
+						updateYield();
```
dguenms: I think this should happen in `setCurrentEra` and only for the Moors as part of their UP.

dguenms (changes requested):

MeuhV1:
Done

dguenms (approved):

PR #341, opened 2020-08-15
Improved Great People 10

etiennefd:
Various additions collected over many months


Assets/Python/GreatPeople.py
```diff
@@ -2341,10 +2374,12 @@ def setup():
 		"fJeanne d'Arc", # 15th
 		iRenaissance,
 		u"Louis de Bourbon-Condé", # 17th
-		"Henri de la Tour d'Auvergne", # 17th
```
dguenms, line 2344: Henri de Turenne

Assets/Python/GreatPeople.py
```diff
@@ -2466,6 +2501,7 @@ def setup():
 		"John Stott", # 20th
 	],
 	iGreatArtist : [
+		u"Ælfric of Eynsham", # 10th
```
dguenms: The game cannot render this character, so it has to be Aelfric

dguenms (changes requested):

Assets/Python/GreatPeople.py
etiennefd: I think the nickname alone is better, since I don't think he is really referred to as "Henri de Turenne" at all. This is similar to Molière, Voltaire, Montesquieu, all of whom are much better known by their nickname/pseudonym than their full name.

Assets/Python/GreatPeople.py
etiennefd: "æ" is present in many Scandinavian names already; are you sure it can't be rendered?

dguenms: Oh, I checked again and it is.

dguenms (approved):

Assets/Python/GreatPeople.py
etiennefd: I removed the Æ before writing my comment, so you might want to put it back in.

PR #342, opened 2020-08-24
Improved Great People 11

etiennefd:
With the recent overhaul to spies, Great Spies will surely become more common. That sounded like the right time to finally get some Byzantine spy names, which is... hard to do, even though Byzantium is known for espionage and surveillance and have a relevant UP. I found several other spy names in the process.

Some of them are pretty obscure, so:
- **Aristo**: name of an agent of Hannibal according to (1)
- **Mihr Narseh, Yazdgushnasp, Fariburz**: powerful officials of the Sasanian empire, which according to (2) must have headed extensive intelligence networks. I wanted to give Persia pre-islamic spies, but maybe they're not relevant.
- **Hassan-i Sabbah**: already included for the Arabs as the founded of the Assassins, but he was Persian
- **Locusta**: famous poisoner. I'm not keen on putting random assassins in the list, but she sounded too notorious not to include
- **Lucius Blassius Nigellio**: only name I could find for a Roman speculator, a class of spies/informants (see (2))
- **Paulus Catena**: a.k.a. Paul the Chain, mentioned in (2)
- **Palladios**: I don't even actually remember, let me check... Oh yeah, in (2): "Despite the general weakness of the Agentes in Rebus after 365, there were some famous and remarkable individuals who made all the difference. The best example of this is the fame of the agens Palladius. Theodosius II (408-450) and the Roman top brass used Palladius as their trusted courier to carry the most important messages in the early 5th century. This man was able to reach the eastern frontier in three days and then return back to the capital in three days."
- Empress **Theodora**: already included as a prophet, but (3) reports that she had her own spy network
- **Stauriakos**: https://en.wikipedia.org/wiki/Staurakios_(eunuch)
- **Samonas**: https://en.wikipedia.org/wiki/Samonas, also discussed in (1)
- **Nikolaos** (Nicholas): name of a former tax collector who fled Byzantium, settled in Bagdad and then became an informant for the Byzantines, according to (1) and (3)
- **Sitt al-Mulk**: Fatimid princess who supposedly had her own spy network

The rest should be easy to check.

(1) http://macedonia.kroraina.com/en/fdois
(2) The Eyes and Ears: The Sasanian and Roman Spies ca. AD 222-450 (PDF)
(3) https://openaccess.leidenuniv.nl/bitstream/handle/1887/73138/30-03-05-Koutrakou.pdf?sequence=1

Unrelatedly, I suggest adding Pablo Escobar as a Colombian great merchant, which I guess might be controversial since he was a criminal. But he became a multibillionaire, and the richest criminal in history, through his illicit activities, so he certainly was great at commerce... Let me know if you want me to remove him.

### DIFF
--- a/Assets/Python/GreatPeople.py
+++ b/Assets/Python/GreatPeople.py
@@ -2504,7 +2504,7 @@ iEngland : {
 		"John Stott", # 20th
 	],
 	iGreatArtist : [
-		u"Aelfric of Eynsham", # 10th
+		u"Ælfric of Eynsham", # 10th
 		"Geoffrey Chaucer", # 14th
 		"Thomas Malory", # 15th
 		iRenaissance,
@@ -4347,6 +4347,7 @@ iColombia : {
 		"Julio Mario Santo Domingo", # 20th
 		u"Carlos Ardila Lülle", # 20th
 		"Luis Carlos Sarmiento Angulo", # 20th
+		"Pablo Escobar", # 20th
 	],
 	iGreatEngineer : [
 		u"Carlos Albán", # 19th
--- a/Assets/Python/GreatPeople.py
+++ b/Assets/Python/GreatPeople.py
@@ -933,7 +933,6 @@ iPersia : {
 		"Tiribazus", # 4th BC
 		"Bagoi", # 4th BC
 		"Tiridat", # 1st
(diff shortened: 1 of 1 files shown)


dguenms (approved):
No, I think all of this is reasonable.

PR #343, opened 2020-09-07
Changed UHV tile info code

MeuhV1:
I separated the generation of the UHV areas to a new function. This allows the plotlist to be generated once and multiple tiles can be checked for being in this list. Previously the plotlist was generated for each tile.

I kept the new function in CvScreensInterface.py, but maybe it would be better if it is moved to RFCutils.py or Victory.py.

### DIFF
--- a/Assets/Python/EntryPoints/CvScreensInterface.py
+++ b/Assets/Python/EntryPoints/CvScreensInterface.py
@@ -906,276 +906,152 @@ def getUHVTileInfo(argsList):
 	
 	plot = gc.getMap().plot(x, y)
 	
+	lUHVAreas = getUHVAreas(iCiv)
+	
+	if lUHVAreas:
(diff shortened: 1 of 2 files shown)


dguenms:
I see the problem this is addressing, not sure if I like the solution though. I am currently redoing the victory code anyway, and had planned to tackle area displays there. I will keep this concern in mind.

MeuhV1:
Okay. I will close the PR then.

It isn't a huge issue. Most of the time this is used for the tile tooltip, which is only 1 tile at a time. (My update is probably even marginally slower) It only effects when you want to check multiple tiles. AFAIK only my UHV area viewer in the WB does that.

I noticed this problem when I tried to view the UHV area of a civ and it took several second to show, while it was almost instantly before the code rewrite.

PR #345, opened 2020-10-01
Fix BUG's unit rename functionality

Alexius08:
1. A bug prevented the random name generator from working properly.
2. The rename functionality hasn't been updated to reflect changes made to the unit roster (e.g. removing corporate executives, splitting mounted units into light and heavy cavalry)

Both have been fixed. Other issues related to the unit rename functionality that need to be fixed are:

1. Fixing unit numbering when units are spawned, when the player either has spawned or got conqueror events. I have trouble finding the events that are triggered when new units are spawned.
2. American Pioneers are somehow counted as gunpowder units. Is there a specific reason for this?


dguenms:
Oh, thanks for doing this. I am not always paying full attention to these BUG features.

For your questions:
1. How does it work right now? via `onUnitTrained`? Units are spawned from Python via the `makeUnits` function in `Core.py`, which uses `CyPlayer.initUnit`. As far as I can tell, this does not trigger a Python event. But we could make `makeUnits` fire its own custom Python event. Look into `Events.py` if you want to try your hand at this yourself, or maybe we can coordinate so I can make this possible.
2. What unit combat type do you expect them to have?

Let me know if you want to continue discussion here. If not, I can merge this and we can keep talking on CFC.

dguenms (approved):

Alexius08:
We can go ahead with the merge. I'll elaborate on what I'm trying to do there.

dguenms (approved):

PR #347, opened 2020-12-28
Unit overhaul

Migck:
My work going through military units

This started when I noticed the mess that is the heavy and light cavalry lines, with stuff like first strike immunities and flanking atack capabilities bouncing back and forth between them. Then I noticed stuff with ships, and some more things here and there, and have tweaked everything until my mind is a bit more at peace.

So, from the beginning of the XML, downwards, more or less:
- Warrior units you can pop from goodie huts now upgrade to Light Swordman instead of Axeman
- Disabled Warrior and Axeman by setting their cost to -1, instead of setting them to national units (0 allowed)
- Some unique units I view them as so useful, they skip an upgrade so they can coexist with it. Starting with the Roman Legion, which quite simply does everything the Heavy Swordman does and more, now upgrades to Musketeer
- Heavy Swordman no longer has the counter AI types, a leftover from its vanilla +50% vs. Melee Units.
- It now can only be made with Iron instead of Copper, relying on bronze is not possible if you want to have effective and plentiful weapons and armor for melee
- I made an exception for the Viking Huscarl because there's no Iron in Scandinavia
- Persian Immortal skips Heavy Spearman and upgrades to Pikeman
- Changed default unit AI of Arquebusier from attack to city defense, it's another leftover from the vanilla version which doesn't upgrade from other units and coexisted with crossbowmen. And it's supposed to fill that role when spawned on colonizing galleons
- Gave option for collateral AI on Chinese ChuKoNu and Firelancer, and British Redcoat
- Changed default AI of Grenadier from counter to collateral, it's better for them to be used aggressively and they don't really counter that much
- All versions of it now have the Tamil Rocketeer's +25% attack vs Musketeer as well as Rifleman, since they can coexist with both, and they now upgrade to Sapper instead of Infantry
- Anti-tank renamed to Sapper, it has features of the Grenadier: 50% withdraw chance, collateral damage, and city raider I promotion, as well as their old features. Its +100% vs. Armored units is changed to +100% on attack only, and it can enter jungles and rainforest. It doesn't upgrade further, remaining as a cheap and annoying "guerrilla warfare" unit. This probably should be evaluated in a different commit, sorry, I am not good with computer
- Mobile SAM defends first against Gunships, way I see it a guy carrying a shoulder mounted weapon (SAM Infantry) would have a hard time anticipating aerial attacks but a hi-tech mobile platform would have no trouble with that
- Disabled the Paratrooper, and instead made the Marine take over its functions. It requires Radar and Social Services instead of Globalism, and targets Siege Weapons first outside cities. Again, another change which likely belongs in its own commit
(BTW, with Globalism beelined, Marines could be made without Ballistics. There are also a few units which likely use high caliber guns that could be made without it, like the Tank, Cruiser, or Battleship. Not sure if that's worth correcting for realism's sake)
- Gave Nubian Medjay attack default AI, but maybe it's not needed

The Cavalry is here:
It seems there was a bit of confusion from the beginning, since Chariot has both Horseman and Horse Archer in two separated unit upgrade blocks.
The way I've set it up, the Light Cavalry line is the one that gets the first strike immunity, open terrain attack, and Siege Weapon combat bonus and flanking. The Heavy Cavalry line only has its raw strength and mobility to work with.
- Chariot is Light Cavalry and upgrades to Horse Archer, has +25% against siege and flanking against Catapults
- Horseman is heavy Cavalry and upgrades to Lancer, has no flanking nor bonus against Siege.
- Greek Companion cavalry has attack city default AI since they have no penalty to it
- Camel Rider is Light Cavalry like later upgrades
- Horse Archer begins the immunity to first strikes. I also gave it flanking against Bombard, at half strength
- Camel Archer also has bonus and flanking against siege
- Indian varu has -25% against Siege, this thing is monstrous like the vanilla Cataphract
- Mongolian Keshik is a bit special since it is Light Cavalry replacing a Heavy unit, I've just given it collateral possible AI
- Pistolier has +25% against siege to be on par with the other Light Cavalry units
- Spanish Conquistador can no longer enter Marshes, it was the only unit for most of the game capable of doing that and there aren't any in Spain's historical area anyways
- Colombian Llanero is now Light Cavalry instead of Heavy, it replaces the Hussar and I don't see anything "heavy" about them. They can now enter Jungle and Rainforest, of which there are plenty in northern South America
- Cavalry unit and Mexican Rurales is where both the Heavy and Light lines end, so they end up with both their features. Namely they are also immune to first strikes
- Khemer Ballista Elephant has city attack AI and 1 first strike, I think otherwise the "Ballista" part doesn't really do much for them
- Gunship is also immune to first strikes, and now also targets Armored and Siege units first. Without SAM Infantry to hunt them down or Mobile SAMs to defend against them, they should easily be able to tear down any combat materiel they come across

Back to non cavalry:
- Roman Ballista skips Trebuchet and upgrades to Bombard, it is plain stronger when used in the field
- Trebuchet can now use the same AI types as the Catapult, a vanilla leftover where those were forbidden to it to force it into a city siege role, since it was weaker than the catapult when attacking out in the field
- Mughal Siege Elephant can enter Jungle and Marsh, but is not immune to siege weapons' collateral damage. I don't know, did these things even exist? How do you reload a cannon atop an elephant?
- Howitzer and Mobile Artillery don't require Iron, like other modern units for which the availability of iron is supposed to no longer be a factor

And now the Navy:
There is now a "Transport" line with the... transport ships, separated from a "Warship" line with most of the other ships. With a few UU exceptions, otherwise the upgrade lines were making too many weird jumps and changes in function for my tastes
- Forbade Worker Boats from exploring AI because I'm sick of the AI using them for that
- Forbade Polynesians Wakas from the Assault AI, may not be necessary since Polynesia is not even supposed to be AI enabled
- War Galley has 50% withdraw chance, Heavy Galley and most/all following in the warship line have 25%
- War Galley no longer upgrades to Caravel, only to Galleass, the regular version that is
- Byzantine Dromon upgrades to Ship of the Line, I think they make a good companion of the Galleass with their collateral damage, and AI Byzantium is likely dead before they can even fleet Frigates
- Viking Longship is so plainly superior to the Cog that I had to make up some sort of handicap for it. It can now only transport units of the Swordman line (light, regular, heavy, and Huscarls), and upgrades to Galleon instead of Galleass. So they can be a tad anachronistic but hey, early water bombardment
- Cog no longer upgrades to Caravel, only to galleon, and requires Shipbuilding
- Galleass upgrades to Frigate instead of Torpedo Boat, and also requires Shipbuilding
- Caravel, Torpedo Boat, and Privateer thus don't come from previous units and can remain existing together with even Ironclads: Caravels can explore rival territory and drop off spies and such, Privateers become outclassed by everything but hey, a pirate's life up to even 1850 or so was about not getting caught. They all upgrade to Submarine
- Carrack default AI is settler sea, hopefully that incentivizes the Portuguese to colonize early
- East Indiaman skips Steamship, since it is so strong, has 1 more cargo and doesn't need open borders I imagine it'd be useful up to diesel engines
- Ship of the Line and English Man o' War upgrade to Ironclad, and have 1 less movement (5 and 6), Frigate has 6 and I think it makes sense for them to be slower.
- Ironclad has 25% withdraw like the Frigate and Cruiser, and 7 movement like a Steamship
- Torpedo Boat, Submarine and Attack Submarine now target the transport line first. Destroyer and Stealth Destroyer can counter this by defending against them first
- Cruiser upgrades to either Battleship or Destroyer, has 8 movement and collateral damage like the Ironclad
- Canadian Corvette is so fast (13 movement, fastest in fact) and seems so specialized, that I felt it shouldn't upgrade to Stealth Destroyer, and can work alongside it
- Stealth Destroyer has 30% air interception like Destroyer and Missile Cruiser
- Submarine can carry 1 people like Caravel and Nuclear Submarine. Both submarines also have +50% attack vs Ironclads, Cruisers and Battleships, considering those should have a very hard time defending against them

And the finale
- Great Spies have 3 moves, and the Sentry and Logistics II promotions (+1 sight, -1 terrain movement cost and can use enemy roads). I wanted them to be capable of doing reconnaissance like a top level spy unit (those would use Security instead of Sentry for the +1 sight)

This is an awful lot, take what you will or don't, hopefully something catches your attention
Of course there would also be parts of the pedia and pedia strategy texts that ought to be updated, but that's hardly worth it IMO

### DIFF
--- a/Assets/XML/Text/Units.xml
+++ b/Assets/XML/Text/Units.xml
@@ -3557,4 +3557,21 @@
 		<Italian></Italian>
 		<Spanish></Spanish>
 	</TEXT>
+	
+	<TEXT>
+		<Tag>TXT_KEY_SPECIALUNIT_SEARAIDER</Tag>
+		<English>Swordmen</English>
+		<French>Swordmen</French>
+		<German>Swordmen</German>
+		<Italian>Swordmen</Italian>
+		<Spanish>Swordmen</Spanish>
+	</TEXT>
+	<TEXT>
+		<Tag>TXT_KEY_UNIT_AT_INFANTRY</Tag>
+		<English>Sapper</English>
+		<French>Sapper</French>
+		<German>Sapper</German>
+		<Italian>Sapper</Italian>
+		<Spanish>Sapper</Spanish>
+	</TEXT>
 </Civ4GameText>
--- a/Assets/XML/Units/CIV4SpecialUnitInfos.xml
+++ b/Assets/XML/Units/CIV4SpecialUnitInfos.xml
@@ -35,10 +35,6 @@
 					<UnitAIType>UNITAI_MISSIONARY_SEA</UnitAIType>
 					<bUnitAI>1</bUnitAI>
 				</CarrierUnitAI>
-				<CarrierUnitAI>
-					<UnitAIType>UNITAI_ASSAULT_SEA</UnitAIType>
-					<bUnitAI>1</bUnitAI>
-				</CarrierUnitAI>
 			</CarrierUnitAIs>
 			<ProductionTraits/>
 		</SpecialUnitInfo>
@@ -77,6 +73,20 @@
 			<bPlayerValid>0</bPlayerValid>
 			<bCityLoad>1</bCityLoad>
 			<CarrierUnitAIs/>
+			<ProductionTraits/>
+		</SpecialUnitInfo>
+		<SpecialUnitInfo>
+			<Type>SPECIALUNIT_SEARAIDER</Type>
+			<Description>TXT_KEY_SPECIALUNIT_SEARAIDER</Description>
+			<bValid>1</bValid>
+			<bPlayerValid>1</bPlayerValid>
+			<bCityLoad>0</bCityLoad>
+			<CarrierUnitAIs>
+				<CarrierUnitAI>
+					<UnitAIType>UNITAI_ASSAULT_SEA</UnitAIType>
+					<bUnitAI>1</bUnitAI>
+				</CarrierUnitAI>
+			</CarrierUnitAIs>
 			<ProductionTraits/>
 		</SpecialUnitInfo>
 	</SpecialUnitInfos>
--- a/Assets/XML/Units/CIV4UnitClassInfos.xml
+++ b/Assets/XML/Units/CIV4UnitClassInfos.xml
@@ -179,7 +179,7 @@
 			<Description>TXT_KEY_UNIT_WARRIOR</Description>
 			<iMaxGlobalInstances>-1</iMaxGlobalInstances>
 			<iMaxTeamInstances>-1</iMaxTeamInstances>
-			<iMaxPlayerInstances>0</iMaxPlayerInstances>
+			<iMaxPlayerInstances>-1</iMaxPlayerInstances>
 			<DefaultUnit>UNIT_WARRIOR</DefaultUnit>
 		</UnitClassInfo>
 		<UnitClassInfo>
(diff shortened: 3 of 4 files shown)


dguenms:
Incorporated some of the suggested changes, thanks.

PR #352, opened 2021-11-24
Add new Rise implementation

dguenms:


LT-Rascek:
Potential bug:  Arab conquerors spawning circa 1174 AD to conquer Tarabulus; I don't see any mention of Arab conquerors in Assets/Python/AIWars.py for that period. Attached save is 1 turn before spawn.

[RFC_DOC_Phonecia_ArabTrabulusConquerors.zip](https://github.com/dguenms/Dawn-of-Civilization/files/7614869/RFC_DOC_Phonecia_ArabTrabulusConquerors.zip)

Another potential issue: Babylon will flip on Arab spawn even if Human player is Babylon and is the only human city. Similar issue with Phoenicia and Moors; you can loose your whole core if you don't settle Oea as well.

PR #366, opened 2024-06-12
Fix Norse settler and war maps

Alexius08:
The stability files for the Norse have wrong filenames. Renaming them allows them to be read, although this works only for new games.


Alexius08:
Fix done with other additions on develop: https://github.com/dguenms/Dawn-of-Civilization/commit/7672c1349624655ad59fa369af860d105e8bace7

PR #378, opened 2025-10-19
BUGFIX: wrong cores and capitals returned

bugsnotabunny:
To reproduce the bug before this commit:
1) start 600AD scenario as Norse
2) In python console from Core import *
3) In python console print plots.capital(11) (11 is Celts)

Should return cords of Dublin, but instead returns cords of Paris since player.getPeriod() method always returns -1 instead of set period


dguenms:
Functions like this are type aware. `11` refers to the player slot 11. `Civ(11)` which is the actual value of `iCelts` refers to the civilization.

PR #384, opened 2026-07-10
Improve performance of CvGame::setUpdatePlotGroups by only updating plots with a flag set

OscarBrink:
### Description
This is an attempt to improve performance at end of turn when `CvGame::setUpdatePlotGroups()` is called. It is related to recent discussions in the "Bug Reports and Technical Issues" thread on civfanatics.com.

`CvGame::setUpdatePlotGroups()` takes a long time running through all of the plots in the map when `bNewValue` is `true`. I verified this by doing some simple timings in the code using the >20min America save linked in the thread (https://forums.civfanatics.com/threads/bug-reports-and-technical-issues.533867/post-16969667) I measured two calls to `CvGame::setUpdatePlotGroups` taking approximately 17 and 5 minutes on my computer.

### Fix
Added a `m_shouldUpdatePlotGroup` flag to `CvPlot` which is set to `true` when `updatePlotGroup()` is called and `m_bUpdatePlotGroups` is set to `false` in `CvGame`. Call it a "dirty bit".

Later when `CvGame::setUpdatePlotGroups()` is called with `bNewValue` it checks the `m_shouldUpdatePlotGroup` and only calls `updatePlotGroup` on the plots that have the flag set.

After this change both of the America saves linked in the thread take <1 min to complete on my computer.

### Misc
The flag is set to `false` at the end of `CvPlot::updatePlotGroup(PlayerTypes ePlayer, bool bRecalculate)`. Maybe this should be done in `CvGame::setUpdatePlotGroups()`?

Also not sure if the naming of the functions/variables are good.

I'm not very familiar with this code base, so I don't know if these changes have any unintended consequences.

### DIFF
--- a/CvGameCoreDLL/CvGame.cpp
+++ b/CvGameCoreDLL/CvGame.cpp
@@ -10953,7 +10953,11 @@ void CvGame::setUpdatePlotGroups(bool bNewValue)
 	{
 		for (int iI = 0; iI < GC.getMap().numPlots(); iI++)
 		{
-			GC.getMap().plotByIndex(iI)->updatePlotGroup();
+			CvPlot *plot = GC.getMap().plotByIndex(iI);
+			if (plot->shouldUpdatePlotGroup())
+			{
+				plot->updatePlotGroup();
+			}
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


OscarBrink:
I should note that if I have misunderstood the code and `CvGame::setUpdatePlotGroups` should _always_ update _all_ plots, then this PR will not work.

PR #385, opened 2026-07-16
Fix use after free bug in CvCity::kill

OscarBrink:
### Bug
`CvCity::kill` calls `CvPlayer::deleteCity` which removes the `CvCity` from a list in the `CvPlayer`, but also calls delete on this `CvCity`. When `getOwnerINLINE()` is called after that it can return some bogus value if the memory has been overwritten.

```cpp
void CvCity::kill(bool bUpdatePlotGroups)
{
 	...
	GET_PLAYER(getOwnerINLINE()).deleteCity(getID());  // <-- this is deleted here
	GET_PLAYER(getOwnerINLINE()).updateCultureRanks(); // getOwner() then used on deleted CvCity
	...
```

### Fix
Use the `eOwner` value which was set `eOwner = getOwnerINLINE()` earlier in the function.

### Misc
I have a save where this causes a crash for me on end turn. Let me know if you need it.

### DIFF
--- a/Assets/Python/Rise.py
+++ b/Assets/Python/Rise.py
@@ -703,9 +703,10 @@ class Birth(object):
 					capital = completeCityFlip(city, self.iPlayer, city.getOwner(), 100)
 				else:
 					self.data.lPreservedWonders += [iWonder for iWonder in infos.buildings() if isWonder(iWonder) and city.isHasRealBuilding(iWonder)]
-				
-					plot_(city).eraseAIDevelopment()
-					plot_(city).setImprovementType(iCityRuins)
+
+					cityPlot = plot_(city)
+					cityPlot.eraseAIDevelopment()
+					cityPlot.setImprovementType(iCityRuins)
 		
 		if capital:
 			self.prepareCity(capital)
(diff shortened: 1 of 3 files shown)


OscarBrink:
Got another use after free when Prussia was spawning and Magdeburg was being razed:

### Bug
`CvPlot::eraseAIDevelopment()` is called which calls `CvCity::kill` if the plot has a city on it. Then the `CvCity` is used to set the plot to ruins.

```py
	def prepareCapital(self):
		...
		if self.iCiv not in lInvasionCivs:
			for city in cities.ring(self.location):
				...
				else:
					...
					plot_(city).eraseAIDevelopment()           # <-- Calls CvCity::kill on city
					plot_(city).setImprovementType(iCityRuins) # <-- plot_(city) Use after free
```

### Fix
Only call `plot_(city)` once and use the value for both `eraseAIDevelopment` and `setImprovementType`.

#### Side Note
I am running the game on Linux under wine, which is why I believe I am being hit by relatively many of these issues.

My best guess (although I haven't investigated it) is that there is some difference with the underlying allocator implementation on Wine/Linux and on Windows that causes this.

So maybe these issues never actually caused any problems in practice on Windows.

OscarBrink:
Also rebased on top of latest develop

dguenms (approved):

